PR #55, opened 2026-04-16
Dynamic MVT tile endpoint for H3 hex visualization

cboettig:
## Summary

- Adds `register_hex_tiles` MCP tool that materializes a pre-aggregated H3 resolution pyramid to S3 and returns a MapLibre-compatible vector tile URL template
- Adds `/tiles/hex/{hash}/{z}/{x}/{y}.pbf` endpoint serving MVT tiles from the pyramid via DuckDB `ST_AsMVT`
- New `tiles/` package: `tile_math.py` (pure helpers), `db.py` (persistent DuckDB connection), `pyramid.py` (pyramid materialization), `endpoint.py` (Starlette request handler)
- Docs: `h3-guide.md` usage section, `assistant-role.md` dispatch hints, R & Python programmatic access examples

Implements #4.

## Test plan

- [ ] 32 unit/integration tests pass locally (`test_tile_math`, `test_tile_pyramid`, `test_tile_endpoint`, `test_server::TestTileRouteMounted`)
- [ ] Deploy test server from this branch and register a small tileset against real S3
- [ ] Verify tile responses decode correctly in a MapLibre client
- [ ] Check no regression in existing `query` tool and STAC endpoints

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,21 @@ Examples:
 
 
 
-### Using Claude Code (Desktop)
+### Using Claude Code (CLI)
+
+Run this command once in your terminal:
+
+```bash
+claude mcp add --transport http duckdb-geo https://duckdb-mcp.nrp-nautilus.io/mcp
+```
+
+To make it available across all your projects, add `--scope user`:
+
+```bash
+claude mcp add --transport http --scope user duckdb-geo https://duckdb-mcp.nrp-nautilus.io/mcp
+```
+
+### Using Claude Desktop
 
 Add to your Claude Desktop configuration file:
 
--- a/assistant-role.md
+++ b/assistant-role.md
@@ -2,3 +2,14 @@ You are a geospatial data analyst with expertise in cloud-native geospatial form
 
 You help users query and analyze geospatial datasets stored as Parquet files on S3 object storage.
 Use the available tools to discover datasets, run SQL queries, and interpret results.
+
+### Dispatch: map-rendering tools
+
+- `query` — for answering in markdown tables (capped at 50 rows). Default.
+- `register_hex_tiles` — when the user wants *hexes on the map*, or the result
+  would be a large (>100k cell) hex layer. Returns a tile URL; the client adds
+  it as a MapLibre vector source.
+
+If the user asks to "show", "render", or "color" hexes on the map, choose
+`register_hex_tiles`. If they ask "what's the value of X at location Y" or
+"top N hexes by Z", use `query`.
--- a/docs/.vitepress/config.js
+++ b/docs/.vitepress/config.js
@@ -17,6 +17,7 @@ export default {
           { text: 'Quick Start', link: '/guide/quickstart' },
           { text: 'Available Datasets', link: '/guide/datasets' },
           { text: 'Private Data Access', link: '/guide/private-data' },
+          { text: 'Programmatic Access (R & Python)', link: '/guide/programmatic-access' },
         ],
       },
       {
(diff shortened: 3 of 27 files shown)


cboettig (commented):
## Code review findings — items to address before merge

### 1. `_read_metadata` creates a new DuckDB connection per tile request (`tiles/endpoint.py:44-64`)

Every tile request opens a fresh `:memory:` DuckDB connection just to read the metadata sidecar. The persistent `tile_con` (already loaded with `httpfs`) is available on `request.app.state` and should be reused, or the metadata should be cached on `app.state` after the first read (it never changes post-registration).

### 2. `agg` parameter is not validated (`tiles/pyramid.py:81`)

`agg` is interpolated directly into the generated SQL (`{agg}({col})`). An unexpected value produces a DuckDB syntax error (not silent injection), but a one-line allowlist check is cheap insurance:
```python
VALID_AGG = {"AVG", "SUM", "MIN", "MAX", "COUNT"}
if agg.upper() not in VALID_AGG:
    raise ValueError(f"agg must be one of {VALID_AGG}, got {agg!r}")
```

### 3. Column names are not quoted in `build_pyramid_sql` (`tiles/pyramid.py:29-43`)

`value_columns` and `h3_column` come from user SQL and are interpolated bare into the generated SQL. A column named `my value` or one that shadows a keyword will break the query. Wrap them with double quotes:
```python
value_list_raw = ", ".join(f'"{c}"' for c in value_columns)
value_list_agg = ", ".join(f'{agg}("{c}") AS "{c}"' for c in value_columns)
# and h3_column: f'"{h3_column}"'
```

### 4. `tile_lifespan` is dead code (`tiles/db.py:26-39`)

`tile_lifespan` is defined but never wired into `server.py` — the server uses the global `_get_tile_con()` lazy-init instead, so there is no clean shutdown of the persistent connection. Either use `tile_lifespan` in `server.py` (replacing the global), or delete it.

### 5. `app.router.routes.append()` is non-standard (`server.py:255`)

Starlette's `app.add_route()` is the public API for adding routes after construction. The direct list append works today but bypasses internal router state and could silently break on a Starlette version bump. Prefer:
```python
app.add_route("/tiles/{namespace}/{name}/{z:int}/{x:int}/{y:int}.pbf", serve_tile)
```

cboettig:
## Code review findings — items to address before merge

### 1. `_read_metadata` creates a new DuckDB connection per tile request (`tiles/endpoint.py:44-64`)

Every tile request opens a fresh `:memory:` DuckDB connection just to read the metadata sidecar. The persistent `tile_con` (already loaded with `httpfs`) is available on `request.app.state` and should be reused, or the metadata should be cached on `app.state` after the first read (it never changes post-registration).

### 2. `agg` parameter is not validated (`tiles/pyramid.py:81`)

`agg` is interpolated directly into the generated SQL (`{agg}({col})`). A one-line allowlist check is cheap insurance:
```python
VALID_AGG = {"AVG", "SUM", "MIN", "MAX", "COUNT"}
if agg.upper() not in VALID_AGG:
    raise ValueError(f"agg must be one of {VALID_AGG}, got {agg!r}")
```

### 3. Column names are not quoted in `build_pyramid_sql` (`tiles/pyramid.py:29-43`)

`value_columns` and `h3_column` come from user SQL and are interpolated bare. A column named `my value` or one that shadows a keyword will break the query. Wrap them in double quotes:
```python
value_list_raw = ", ".join(f'"{`c`}"' for c in value_columns)
value_list_agg = ", ".join(f'{agg}("{`c`}") AS "{`c`}"' for c in value_columns)
```

### 4. `tile_lifespan` is dead code (`tiles/db.py:26-39`)

Defined but never wired into `server.py` — the server uses the global `_get_tile_con()` lazy-init instead, with no clean shutdown. Either wire `tile_lifespan` or delete it.

### 5. `app.router.routes.append()` is non-standard (`server.py:255`)

Prefer `app.add_route()` which is the public Starlette API and won't silently break on a version bump.

cboettig:
Superseded by #60 which merged all dev work (tiles + get_collection + S3 fix) into main.

Issue #56, opened 2026-04-16
Add get_collection tool returning structured STAC JSON


cboettig:
## Context

The geo-agent app (`boettiger-lab/geo-agent`) currently fetches STAC collections directly from the browser to build map layers and LLM context. This creates two problems:

1. **Redundant STAC parsing** — the geo-agent reimplements STAC schema/path extraction that the MCP server already does (and does better — see boettiger-lab/geo-agent#163, boettiger-lab/geo-agent#162).
2. **Divergent implementations** — the geo-agent's `extractColumns()` reads only collection-level `table:columns`, missing per-asset schemas. The MCP's `_extract_parquet_assets()` correctly inlines per-asset `table:columns`. This caused real failures in production (model got wrong column set, guessed paths, wasted round-trips).

The broader goal is to centralize STAC access behind the MCP server: the MCP handles data access (fetching, caching, serving STAC metadata), and the geo-agent handles only presentation (map layers, LLM orchestration).

## Proposal

Add a `get_collection` tool that returns the STAC collection as structured JSON — not the markdown text that `get_stac_details` produces for LLM consumption, but the actual collection data suitable for programmatic use.

### What it returns

The full STAC collection object (or a lightly normalized version), including fields that `get_stac_details` intentionally omits because they're irrelevant to SQL queries:

- **All assets** — parquet, PMTiles, COG, GeoJSON (not just parquet)
- **Per-asset fields** — `href`, `type` (media type), `title`, `description`
- **STAC extensions** — `vector:layers`, `raster:bands` (nodata, `classification:classes`, color hints), `table:columns` (both collection-level and asset-level), `file:size`
- **Collection metadata** — `id`, `title`, `description`, `providers`, `license`, `extent`, `keywords`, `links`, `summaries`
- **Child collection IDs** — for parent containers like `wyoming-wildlife-lands`
- **S3 path conversion** — `href` values converted from HTTPS to `s3://` form (reusing `_href_to_s3()`)

### Why not extend `get_stac_details`?

The two tools serve different consumers:

| | `get_stac_details` | `get_collection` |
|---|---|---|
| **Consumer** | LLM (reads text) | App code (reads JSON) |
| **Format** | Markdown text | Structured JSON |
| **Assets** | Parquet only | All (PMTiles, COG, GeoJSON, parquet) |
| **Schema** | Formatted column descriptions | Raw `table:columns` arrays |
[...]

PR #57, opened 2026-04-16
feat: add get_collection tool returning structured STAC JSON

cboettig:
## Summary

Implements #56.

- **`stac.py`**: adds `_collection_to_dict()` that converts a `pystac.Collection` to a structured dict — all assets (parquet, PMTiles, COG, GeoJSON), per-asset STAC extension fields (`table:columns`, `raster:bands`, `vector:layers`), `href` values converted from HTTPS to `s3://`, plus full collection metadata (providers, extent, license, keywords, external links, child IDs).
- **`stac.py`**: populates a parallel `_STAC_RAW` cache alongside `STAC_DATASETS` during startup — zero extra network calls, same `dict` lookup performance as `get_stac_details`.
- **`stac.py`**: adds `get_collection()` with the same lookup / fuzzy-match / cache-refresh pattern as `get_dataset()`, plus on-demand fetch for non-default `catalog_url`.
- **`server.py`**: registers `get_collection` as an MCP tool.
- **`tests/test_stac.py`**: 20 new tests covering `_collection_to_dict` (href conversion, asset extra_fields, children, nav-link filtering, empty assets, ISO extent serialization) and `get_collection` (dict return, error cases, child exposure, MCP tool registration).

## Difference from `get_stac_details`

| | `get_stac_details` | `get_collection` |
|---|---|---|
| **Consumer** | LLM (reads text) | App code (reads JSON) |
| **Format** | Markdown | Structured dict |
| **Assets** | Parquet only | All types |
| **Schema** | Formatted descriptions | Raw `table:columns` arrays |
| **Purpose** | Help LLM write SQL | Help app build map layers + system prompt |

## Test plan
- [x] 62 tests passing locally (`test_stac`, `test_tile_math`, `test_tile_pyramid`, `test_tile_endpoint`, `test_server::TestTileRouteMounted`)
- [ ] Connect from geo-agent and call `get_collection("wdpa")` — verify all PMTiles + parquet assets present
- [ ] Verify `children` field populated for a parent collection (e.g. `wyoming-wildlife-lands`)

### DIFF
--- a/server.py
+++ b/server.py
@@ -11,7 +11,7 @@ from mcp.server.transport_security import TransportSecuritySettings
 from mcp.shared.session import BaseSession
 from starlette.middleware.base import BaseHTTPMiddleware
 from starlette.responses import JSONResponse
-from stac import STAC_DATASETS, STAC_CATALOG_URL, list_datasets as _stac_list, get_dataset as _stac_get
+from stac import STAC_DATASETS, STAC_CATALOG_URL, list_datasets as _stac_list, get_dataset as _stac_get, get_collection as _stac_get_collection
 
 # Workaround for https://github.com/boettiger-lab/mcp-data-server/issues/5
 # send_notification crashes with ClosedResourceError when the client disconnects
@@ -142,6 +142,18 @@ def get_stac_details(dataset_id: str, catalog_url: str = None, catalog_token: st
 
 def get_dataset_details(dataset_id: str) -> str:
     return _stac_get(dataset_id)
+
+@mcp.tool()
+def get_collection(collection_id: str, catalog_url: str = None, catalog_token: str = None) -> dict:
+    """Return structured STAC collection metadata as JSON for programmatic use.
+
+    Unlike get_stac_details (markdown for LLM consumption), this returns the
+    raw collection dict with all assets (parquet, PMTiles, COG, GeoJSON),
+    per-asset STAC extension fields (table:columns, raster:bands, vector:layers),
+    full collection metadata, and child collection IDs. S3 paths are pre-resolved.
+
+    Intended for app code that builds map layers and system prompts programmatically."""
+    return _stac_get_collection(collection_id, catalog_url, catalog_token)
 
 # -------------------------------------------------------------------------
 # 7. MCP PROMPTS (Personas for Smart Clients)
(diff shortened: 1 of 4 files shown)


cboettig:
Addressed all four code-review items from the review:

- **Bug fix**: `test_nav_links_excluded` — nav links were created in a loop but never collected into a list, so `col.links` only ever contained the documentation link and the filtering assertion was vacuous. Fixed by accumulating into `nav_links` and setting `col.links = nav_links + [doc_link]`.
- **`_NAV_RELS` moved to module level** — was being re-created as a new set on every `_collection_to_dict` call; now a module-level constant alongside the other `_S3_*` constants.
- **Removed unnecessary `global _STAC_RAW`** — the declaration was present in `fetch_stac_catalog` but `global` is only needed for rebinding, not for `.update()` mutation.
- **Added comment on cache-miss re-fetch side effect** in `get_collection` to make the `_STAC_RAW` dependency explicit.

cboettig:
## Code review — issues to address

### 1. Full-catalog fetch for single-collection lookup (perf)

Both `get_collection()` and `get_dataset()` iterate the **entire** catalog when `catalog_url` is provided, even though only one collection is needed. With 80+ collections this is wasteful — every call fetches and parses the full tree. Should fetch the target collection directly by ID using `pystac.Collection.from_file()` or similar, falling back to full iteration only when direct access fails.

### 2. Fuzzy match returns first partial hit, not best match

`if collection_id.lower() in key.lower()` returns the first dict-iteration match. Passing `"census"` returns whichever census collection happens to appear first. Should prefer exact match → exact-prefix match → substring, or at least document the behavior.

### 3. `_collection_to_dict` doesn't strip `None` at top level

Asset dicts get `{k: v ... if v is not None}` filtering, but top-level fields (`license`, `description`) can be `None` in the output. Minor inconsistency.

### 4. Missing test coverage for fuzzy match and cache-miss re-fetch

The fuzzy-match path and cache-miss re-fetch path in `get_collection()` have no test coverage.

### 5. `get_dataset_details` is not registered as an MCP tool

`get_dataset_details()` (server.py:143) is a bare function — no `@mcp.tool()` decorator. The docstring on `get_stac_details` tells clients to "use the local get_dataset_details tool instead" for pre-loaded data, but it's unreachable via MCP. Should either register it as a tool or remove the dead code and the misleading docstring reference.

---

Will address these on the branch now.

PR #64, opened 2026-04-16
Add optional query_backend plugin hook

cboettig:
## Summary

- Adds a `try: import query_backend` block near the top of `server.py`
- If `query_backend.py` is present alongside `server.py`, its `execute()` function replaces the built-in DuckDB isolation engine for the `query` tool
- Falls back silently to DuckDB when no plugin is found — existing deployments are unaffected

## Interface contract

`query_backend.execute(sql, s3_key=None, s3_secret=None, s3_endpoint=None, s3_scope=None) -> str`

Plugins receive all credential kwargs; they may use or ignore them.

## Motivation

Enables `mcp-gpu-data-server` (https://github.com/boettiger-lab/mcp-gpu-data-server) to stay in sync with this repo rather than maintaining a diverging fork. That repo's Dockerfile clones this repo at build time and drops in a GPU-accelerated `query_backend.py` via this hook.

## Test plan
- [ ] Server starts normally with no `query_backend.py` present (DuckDB path)
- [ ] Server routes queries through plugin when `query_backend.py` is present

### DIFF
--- a/server.py
+++ b/server.py
@@ -194,7 +194,10 @@ def query(sql_query: str, s3_key: str = None, s3_secret: str = None, s3_endpoint
     """Placeholder (overwritten below)."""
     print(f"🔍 Executing: {sql_query}", file=sys.stderr)
     if _CUSTOM_BACKEND:
-        return _query_backend.execute(sql_query, s3_key=s3_key, s3_secret=s3_secret, s3_endpoint=s3_endpoint, s3_scope=s3_scope)
+        try:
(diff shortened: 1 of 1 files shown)


cboettig:
Superseded by #227. The thin `query_backend.py` import hook here was never adopted — the GPU repo became a full hard fork instead, which is exactly the divergence we want to end. #227 proposes folding the GPU query path into this repo behind an explicit `QUERY_ENGINE` backend abstraction (off by default, zero change to existing deployments), which also handles the real coupling (SQL dialect + h3 function surface) that this 13-line hook couldn't.

Issue #157, opened 2026-05-30
One wedged pod takes down ~50% of MCP traffic — readiness probe only checks TCP, not request-readiness


cboettig:
## Symptom

During the trails-matrix test on 2026-05-29, one DuckDB query in a UI session went runaway (~40 cores / 67 GiB on a single pod for several minutes). The user's expectation, given two replicas, was that the second idle replica would absorb the traffic. Instead the app appeared **completely broken** — even after a hard reload.

5 sequential `POST /mcp initialize` probes from outside the cluster during the incident:

```
probe1: HTTP 000  15.001s   ← TIMEOUT
probe2: HTTP 200   0.117s
probe3: HTTP 000  15.002s   ← TIMEOUT
probe4: HTTP 200   0.123s
probe5: HTTP 000  15.002s   ← TIMEOUT
```

Roughly 50% of new TCP connections were hitting the wedged pod and hanging until the client timed out. The other pod's logs show our IP being served normally — but every other request to the public hostname was a dead end.

## Root cause

The current readiness probe (\`k8s/deployment.yaml\` and \`k8s/dev-deployment.yaml\`):

\`\`\`yaml
readinessProbe:
  tcpSocket:
    port: 8000
  initialDelaySeconds: 15
  periodSeconds: 5
  failureThreshold: 5
\`\`\`

That passes as long as the kernel accepts a TCP connection on port 8000 — and it does, regardless of whether the uvicorn process can actually respond. When one query saturates the pod:

- The uvicorn process is alive enough to accept the TCP connection (the OS handles that), but is starved/blocked on writing the response.
- The TCP-only probe sees the socket as healthy and keeps the pod in the Service's endpoints.
- HAProxy's \`balance-algorithm: leastconn\` counts active TCP connections, not request-readiness. The wedged pod looks equivalent to the idle pod (both at 0 active conns from HAProxy's view, because clients have already disconnected and timed out).
- ~50% of new client connections route to the wedged pod and time out.

There is also no \`livenessProbe\` at all, so a permanently wedged pod is never restarted — it just sits there poisoning the load balance.

## Verified architectural details

- Transport is stateless (\`server.py\`: \`FastMCP(..., stateless_http=True)\`). Replicas are interchangeable per request. Just documented in #156.
- \`uvicorn.run(app, ...)\` has no \`workers=\` argument → single uvicorn worker per pod. Sync MCP tool handlers run in the default asyncio executor (a bounded threadpool).
[...]

PR #177, opened 2026-06-10
Run blocking MCP tools off the event loop (Fixes #176)

### LINKED ISSUE #176: Sync MCP tools block the uvicorn event loop (per-pod concurrency ≈ 1)
## Problem

FastMCP (mcp 1.25.0) `await`s coroutine tool functions but runs **synchronous** ones inline on the uvicorn event loop:

```python
# mcp/server/fastmcp/utilities/func_metadata.py  call_fn_with_arg_validation
if fn_is_async:
    return await fn(**arguments_parsed_dict)
else:
    return fn(**arguments_parsed_dict)   # <- on the event loop, no thread offload
```

All three of our tools are sync and do blocking work:

- `query` — DuckDB scan, up to the 300s MCP timeout
- `register_hex_tiles` — DuckDB prepare probes + an inline `future.result(timeout=5)` wait
- `get_hex_tile_status` — a `time.sleep` long-poll loop, up to 60s

So while any one of them runs, the pod's event loop is blocked: `/healthz` stops answering (readiness pulls the pod within ~15s, even for a *legitimate* long query), tile GETs stall, and other MCP requests queue behind it. Effective per-pod concurrency is **1**; with 6 prod replicas the cluster tops out at ~6 concurrent long queries. The fleet only survives today because the readiness probe + `leastconn` accidentally approximate a 6-slot queue — that's an artifact of probe behavior, not a concurrency model.

`serve_tile` already does the right thing (wraps blocking work in `anyio.to_thread.run_sync`); the MCP tools do not.

## Fix

Register the tools as `async` wrappers that run the blocking implementation in a worker thread (`anyio.to_thread.run_sync`), bounded by a per-tool `CapacityLimiter`. Because the offloaded calls can now overlap on the shared read connection `_tile_con`, also isolate them with `.cursor()` (as the tile-serve path already does) and lock-guard the connection's lazy init.

This keeps `/healthz`, tiles, and other requests responsive while a query runs, and turns each pod from a 1-slot worker into a bounded N-slot one.

## Notes / follow-ups

- `get_hex_tile_status` long-polls still hold a worker thread for the wait duration (bounded by `MCP_STATUS_CONCURRENCY`, default 32). Cheap (parked threads), but a later refactor to `anyio.sleep`-based polling would free them entirely.
- Once this lands, the `/healthz` "event-loop starvation as the busy signal" rationale (#157) needs revisiting — starvation will no longer be the proxy for "busy."

To be tested on **dev** before promoting to prod.

cboettig:
Fixes #176.

## What

FastMCP runs **synchronous** tool functions inline on the uvicorn event loop (mcp 1.25.0). All three of our tools are sync and block: `query` (DuckDB scan, ≤300s), `register_hex_tiles` (prepare probes + 5s inline wait), `get_hex_tile_status` (up to 60s long-poll). So one long call froze the pod — `/healthz` unanswered (readiness pulls the pod), tiles stalled, other MCP requests queued. Effective per-pod concurrency was 1.

This registers the three tools as `async` wrappers that run the blocking body in a worker thread (`anyio.to_thread.run_sync`), bounded by per-tool `CapacityLimiter`s. The event loop stays free, so `/healthz` and tiles stay responsive while a query runs, and each pod becomes a bounded N-slot worker.

## How

- `_threaded_tool(sync_fn, limiter)` — async wrapper using `functools.wraps` so FastMCP derives the **identical** tool name + JSON schema from the wrapped signature (verified by test). Limits are env-tunable: `MCP_QUERY_CONCURRENCY` (8), `MCP_REGISTER_CONCURRENCY` (4), `MCP_STATUS_CONCURRENCY` (32).
- Concurrency safety: offloaded `register_hex_tiles` / `get_hex_tile_status` can now overlap on the shared `_tile_con`. They read/write markers through `.cursor()` (independent handle, same DB + extensions — the tile-serve path already does this), and `_get_tile_con()`'s lazy init is double-checked-lock guarded.
- The sync implementation functions are unchanged in behavior and remain directly unit-testable.

## Tests

- New `TestEventLoopOffload`: wrapper preserves name/doc/signature and is a coroutine; event loop stays responsive while the wrapped impl blocks; registered tools are async with unchanged schemas; register/status use an isolated cursor, not the shared connection.
- Full suite green (229 passed locally).

## Rollout

Test on **dev** (`dev-duckdb-mcp`, tracks `:main`) before promoting to prod. Suggested dev check: fire a deliberate multi-minute query and confirm `/healthz` keeps returning 200 and tiles keep serving during it.

## Follow-ups (not in this PR)

- `get_hex_tile_status` long-polls still hold a worker thread for the wait (bounded by `MCP_STATUS_CONCURRENCY`); an `anyio.sleep`-based rewrite would free them.
- Revisit the `/healthz` "starvation = busy" rationale (#157): starvation is no longer the busy signal once tools run off-loop.

### DIFF
--- a/server.py
+++ b/server.py
@@ -1,3 +1,4 @@
+import functools
 import hmac
 import os
 import re
@@ -125,6 +126,35 @@ def get_isolated_db(s3_key: str = None, s3_secret: str = None, s3_endpoint: str 
         conn.close()
 
 # -------------------------------------------------------------------------
+# 4b. EVENT-LOOP OFFLOAD
+# -------------------------------------------------------------------------
+# FastMCP awaits coroutine tool functions but runs *synchronous* ones inline on
+# the uvicorn event loop (mcp 1.25.0 func_metadata.call_fn_with_arg_validation).
+# Our tools make blocking DuckDB / S3 / long-poll calls, so a single long query
+# would otherwise freeze the whole pod: /healthz stops answering (readiness
+# pulls the pod), tile GETs stall, and other MCP requests queue behind it.
+#
+# `_threaded_tool` wraps a blocking implementation as an async tool that runs the
+# work in a worker thread (anyio.to_thread), keeping the event loop free. A
+# per-tool CapacityLimiter bounds how many of each run at once so N concurrent
+# heavy queries can't oversubscribe CPU/memory. functools.wraps copies
+# __name__/__doc__/__annotations__ and sets __wrapped__, so FastMCP derives the
+# identical tool name + JSON schema from the wrapped function's signature.
+_QUERY_LIMITER = anyio.CapacityLimiter(int(os.environ.get("MCP_QUERY_CONCURRENCY", "8")))
+_REGISTER_LIMITER = anyio.CapacityLimiter(int(os.environ.get("MCP_REGISTER_CONCURRENCY", "4")))
+# Status long-polls are mostly idle waits (cheap to hold), so allow many.
+_STATUS_LIMITER = anyio.CapacityLimiter(int(os.environ.get("MCP_STATUS_CONCURRENCY", "32")))
+
+
+def _threaded_tool(sync_fn, limiter):
+    @functools.wraps(sync_fn)
+    async def _wrapper(**kwargs):
+        return await anyio.to_thread.run_sync(
+            functools.partial(sync_fn, **kwargs), limiter=limiter
+        )
+    return _wrapper
+
+# -------------------------------------------------------------------------
 # 5. MCP RESOURCES (Schema Browsing)
 # -------------------------------------------------------------------------
 @mcp.resource("catalog://list")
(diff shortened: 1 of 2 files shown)


cboettig:
Closing in favor of #210.

This PR predates #185, which landed on `main` afterward and independently solved most of what this PR does — by a different route:

- `register_hex_tiles` and `get_hex_tile_status` are already async wrappers offloading to worker threads on `main`.
- The long-poll in `get_hex_tile_status` already uses `await anyio.sleep` on `main` — strictly better than this PR's version, which holds a worker thread for the whole wait (this PR's own "follow-ups" listed that rewrite as not-yet-done).
- register/status already read markers through `.cursor()` on `main`.

So ~80% of this diff is now redundant with #185, and it conflicts with `main` (DIRTY). The one genuinely still-missing piece — offloading the **`query`** tool, the last heavy sync-on-loop call (DuckDB scans up to 300s) — plus the per-tool `CapacityLimiter` hardening, is carried forward in a small focused PR: #210. The double-checked-lock guard on `_get_tile_con()` was dropped as effectively dead code, since `mount_tiles()` already initializes `_tile_con` at startup before any tool runs.

#176 stays open until #210 merges + rolls to prod.

Issue #193, opened 2026-06-17
Adaptive zoom→res: average-density assumption underserves hotspots; cap auto finest_res on large extents


cboettig:
Follow-up to #188 (closed — the seam fix + adaptive zoom→res shipped in #192, working well on dev). Two related serve/build-side refinements surfaced while testing the adaptive mapping on real continental data.

## 1. Adaptive res uses *average* density → hotspot tiles run fat and slow

The adaptive mapping (`zoom_to_h3_res` in `tiles/tile_math.py`) estimates cells-per-tile as `feature_count_finest · 7^(r−finest) / tiles_covering_bounds(z)` — a **uniform-density** assumption. Real data is clumpy, so tiles over a dense region blow past the per-tile budget.

**Observed (dev, irrecoverable-carbon hex, 282M cells at h9 over the Americas):**
- Predicted ~3,760 cells/tile at z=10 (≈ on the 4000 budget).
- Actual z=10 tiles over the **Amazon carbon hotspot**: 260–346 KB, **3–5 seconds each** (`[tile-serve] ... z=10 ... res=9 bytes=346595 3491ms`).

The average is fine; the hotspots are 3–4× over and slow to serve.

**Possible directions (benchmark-driven):**
- Use a density *percentile* (e.g. p90 cells/tile) rather than the mean when picking res — coarsen enough that the hot tiles stay within budget.
- Or store a coarse per-region density summary at registration and pick res locally per tile (more involved).
- Or accept it and lean on the buffer/clip — but 3–5s tiles are a real UX cost.

## 2. Auto-detected `finest_res` is too aggressive on large extents

`finest_res` is auto-detected from the data's H3 resolution. On continental/global extents this picks res 9 and produces enormous pyramids:
- CONUS carbon → `finest_res=9`, **477M cells**, 178s build.
- Americas carbon → `finest_res=9`, **282M cells**, 97s build.

res 9 (~0.74 km² cells) over a continent is far finer than any view needs and drives both build cost and the fat-tile problem above. Consider capping auto-`finest_res` as a function of extent (coarser ceiling for larger bounds), or warning when `feature_count_finest` crosses a threshold.

Note: #1 is serve-side (res selection), #2 straddles build cost (overlaps #189) and feeds #1. Both are downstream of the adaptive mapping landed in #188/#192.

Related: #188 (seam + adaptive res), #189 (registration/query cost), #178 (hex perf).

Issue #189, opened 2026-06-17
Hex registration query optimization: per-species scans of the 4B-row IUCN hex (318s Phase 1), lower() pruning, polygon-vs-hex


cboettig:
## Summary

`register_hex_tiles` Phase 1 (the source scan) is the dominant build cost and the root of the slow/heavy hex builds. Deferred from the #184 work ("get the pipeline solid first") — this is the agent/data-layout track.

## Evidence

Blue-whale IUCN range hex (`SELECT h5, COUNT(DISTINCT _cng_fid) ... WHERE lower(sci_name)='balaenoptera musculus'`): **Phase 1 = 318s**. The asset `s3://public-iucn/iucn-ranges-2025/hex/h0=*/data_0.parquet` is **4.077 B rows / ~26 GB / 122 geography-partitioned (h0) files**. Querying it *by species* can't prune partitions (sorted by geography), and `lower(sci_name)` defeats even row-group stats pruning (measured **3.3×** penalty on a single partition: 1.2s → 0.36s exact-case).

## Levers (roughly by impact)

1. **Don't scan the 4B-row hex for per-species display.** Hex the species' polygon from the per-species GeoParquet (`iucn-ranges-2025.parquet`, keyed by `id_no` — a handful of polygons for one species) instead of full-scanning the global multi-species hex. **Benchmark this on dev** (hex-the-polygon vs extract-from-existing-hex) across a few species of different range sizes — do not assume; the test will say which wins and when.
2. **Drop `lower()`** — exact-case `sci_name = 'Balaenoptera musculus'` re-enables row-group pruning (~3×). It's a model habit; **advise against it in the dataset's STAC layer guidance** (and/or normalize the column). Find where the agent learns to add `lower()`.
3. **Skip pyramid levels below the data floor + fold stats into the COPY** (carryover from #178 follow-ups) — secondary to Phase 1.

## Notes

- The geo-agent system prompt steers the model to `register_hex_tiles` on-the-fly for IUCN species (high-seas) — revisit that guidance alongside (1).
- Per-species on a geography-partitioned global asset is an access-pattern mismatch; the right fix may be a species-keyed/clustered asset or the polygon path, not query tweaks.

Related: #178 (hex perf), #184 (event-loop; pipeline robustness done there).

cboettig:
[...]

Issue #188, opened 2026-06-17
MVT hex tiles: fix tile-edge seam artifacts (buffer + boundary-crossing cells); dial in zoom_offset


cboettig:
## Summary

MVT hex tiles show **tile-edge seam artifacts**: hexes that straddle a tile boundary get sliced, leaving gaps/half-hexes along tile edges (visible as straight clipping lines across an otherwise hex grid). Worst at coarse zoom where hexes are large relative to the tile. Tracked here along with related MVT-rendering tuning so we can "dial in" the vector path.

## Root cause (the seam bug)

In `tiles/endpoint.py` `_build_tile_sql`:

1. **Center-in-tile selection** — we emit only cells whose *center* lat/lng is inside the tile bbox:
   ```sql
   WHERE h3_cell_to_lat(p.h) BETWEEN {south} AND {north}
     AND h3_cell_to_lng(p.h) BETWEEN {west} AND {east}
   ```
   A hex centered in tile A but overlapping tile B is emitted only by A. B never draws its overlapping part.

2. **No buffer in `ST_AsMVTGeom`** — called as `ST_AsMVTGeom(geom, bounds)` with default extent/buffer, so A's copy is hard-clipped exactly at the A/B edge.

Net: the slice of the hex in B is drawn by nobody → seam.

## Fix

Standard MVT practice — emit boundary-crossing features into *both* tiles and give the geometry a buffer so the renderer stitches across the seam:

1. **Select on a buffered bbox**: expand the cell-selection window by ~one hex radius at the target resolution (or select by geometry-intersects-tile rather than center-in-tile) so each edge hex is emitted by every tile it touches. (Keep dedup semantics in mind — at coarse res a hex may appear in several tiles, which is fine for rendering.)
2. **Pass a buffer to `ST_AsMVTGeom`** (e.g. 64–256 MVT units) so the geometry extends past the tile edge and MapLibre clips/stitches seamlessly.

Acceptance: a coarse-zoom hex layer renders as a continuous grid with no straight clipping lines at tile boundaries.

## Related MVT tuning (same "dial-in MVT" effort)

[...]

PR #194, opened 2026-06-17
k8s(prod): promote v0.7.2 by digest (MVT seam fix + adaptive zoom→res: #188)

cboettig:
Promotes **v0.7.2** to production, pinned by digest `sha256:e22ce38f…`. Previous prod: v0.7.1.

Ships the MVT hex tile improvements from #188 / #192. Both are **serve-time** changes that apply to existing tilesets immediately — **no pyramid rebuild or re-registration**.

## What's in v0.7.2

**Tile-edge seams fixed.** Coarse-zoom hex layers showed straight clip lines / half-hexes at tile boundaries. The cause was center-in-tile cell selection (a hex straddling an edge was emitted only by the tile holding its center). Fix widens the cell-selection window by one hex circumradius so every tile a hex touches emits it; `ST_AsMVTGeom`'s buffer (already 256u by default) stitches the duplicates.

**Adaptive zoom→resolution.** Replaced the fixed linear `res = z − zoom_offset` with a feature-count-aware mapping that uses each tileset's registered extent + cell count to pick the finest H3 res keeping a tile near a cell budget (`TILE_TARGET_CELLS_PER_TILE`, default 4000). Bounded data (e.g. CA) now renders res 5–6 hexes at mid-zoom instead of state-sized blobs; global data stays coarser at the same zoom to bound per-tile size. `zoom_offset` retained as a bias knob; pre-existing metadata falls back to the linear mapping.

## Validation
Deployed to dev and validated on the padus app with continental carbon hex: seams eliminated ("huge improvement"), res ladder confirmed (z1→4, z5→5, z8→8, z10→9). Full test suite green (260 tests).

## Rollout (after merge)
```
kubectl apply -f k8s/deployment.yaml
kubectl rollout restart deployment/duckdb-mcp -n biodiversity
```

## Known follow-ups (non-blocking)
- #193 — adaptive res uses average density, so hotspot tiles (e.g. Amazon carbon) run fat/slow at deep zoom; auto-`finest_res` cap on large extents.
- #189 — agent SQL guidance (spatially clip before hexing; avoid `lower()` on hex assets).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/k8s/deployment.yaml
+++ b/k8s/deployment.yaml
@@ -23,7 +23,7 @@ spec:
                     values: ["us-west"]
       containers:
       - name: server
-        image: ghcr.io/boettiger-lab/mcp-data-server:v0.7.1@sha256:256f012e1c683c472c9ca952e4d4ae9d0c5c90c01caf46eb4f2f77c7f5d8c916
+        image: ghcr.io/boettiger-lab/mcp-data-server:v0.7.2@sha256:e22ce38fe0f84a1d43349bd8a972d6a799df7e282a227737ff06d339c3e27879
         imagePullPolicy: IfNotPresent
         env:
         - name: STAC_CATALOG_URL
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,10 @@
 mcp
-duckdb
+# Pinned: the community `h3` extension is built per exact DuckDB version and
+# lags new releases. Unpinned, CI/image builds pull the latest DuckDB (e.g.
+# 1.5.4) for which `INSTALL h3 FROM community` has no candidate, breaking every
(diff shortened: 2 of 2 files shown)


cboettig:
Added `duckdb==1.5.3` pin (commit dc84ff4): CI was red because unpinned DuckDB resolved to 1.5.4, for which the community h3 extension isn't published yet — every tile test errored at `INSTALL h3 FROM community`. The v0.7.2 image itself is unaffected (built from cached 1.5.3 + h3 layers, validated on dev), but the pin fixes CI here and de-risks future image builds. This PR now: (1) promotes v0.7.2 by digest, (2) pins DuckDB.

PR #241, opened 2026-06-24
feat: drop GeoJSON output path; serve hex tilesets as MVT only

### LINKED ISSUE #219: Drop the GeoJSON output path; serve hex tilesets as MVT only (#190)
## Proposal: drop the GeoJSON output path; serve hex tilesets as MVT only

`register_hex_tiles` currently auto-selects between a single-resolution GeoJSON FeatureCollection (small sets) and an MVT pyramid (large). The #190 work showed MVT now dominates on every axis that matters, and maintaining two output formats costs complexity on **both** the server and the geo-agent client. This issue proposes removing the GeoJSON path entirely.

### Why (benchmarks that motivate it)
- **Render quality**: GeoJSON is single-resolution (no pyramid) → shows real hexes only near the data's native zoom; elsewhere each hex drops below a pixel and merges into a blob (user-confirmed: "cannot see actual hexes much of the time"). MVT's adaptive pyramid renders correctly at every zoom.
- **Client load (recurring)**: GeoJSON re-tessellates the *whole* set on the client on *every* view — measured load-to-interactive ~4s @25k, ~10s @100k, ~18s @200k, ~44s @400k on a 4-core client (steady-state pan stays 60fps; the cost is the one-time-per-view tessellation). MVT tessellates only the visible tiles per viewport.
- **Build cost is ~comparable**: GeoJSON's only build advantage was skipping the pyramid (Phase 2), which is now cheap post-#207 (1.5–8.5s even at 22M cells). Measured build: MVT 25k=5.0s / 100k=6.3s / 200k=6.2s vs GeoJSON 25k=3.4s / 100k=5.1s / 200k=6.7s — within ~1–3s, and MVT is *faster* at 200k.
- **MVT first-paint is small + flat**: serving the fitted view loads coarse pyramid tiles (~0.4–0.8 MB, ~1–2s cold / ~1s warm), roughly independent of dataset size.
- **Simplification**: removes server code (`build_geojson_statement`, `geojson_fc_query`, `_put_gzip`, `_s3_internal_write_url`, the `serve_format` routing in `build_hex_tiles`, the GeoJSON branch in `render_recipe`, `_GEOJSON_MAX_FEATURES`, gzip tests) and a client code path + failure mode (geo-agent#252 was the "format=geojson ignored" bug — removing the format removes that class).

The one real GeoJSON benefit — a portable static file usable outside MapLibre (e.g. QGIS) — is already served by the `query` tool (data as tables/files), not the map-display path.

### Gating checks before removal (benchmarks we'd want)
1. **MVT first-paint for *tiny* bounded sets** (e.g. <5k cells): confirm it's acceptable (≤~2s). We measured ~1–2s at 100–200k; tiny should be faster, but verify the smallest case so we don't regress the "quick little map" experience.
2. **geo-agent dependency audit**: confirm nothing consumes the `geojson_url` / `format=="geojson"` contract that MVT (`tile_url_template` / `format=="vector"`) doesn't already cover. geo-agent is the side that had the GeoJSON-handling bug (#252), so removal should simplify it.

### Scope (cross-repo)
- **mcp-data-server**: remove the GeoJSON build + gzip-PUT + routing + serve-format branching + render-recipe branch + cutoff + tests; `register_hex_tiles` always builds the pyramid.
- **geo-agent**: drop GeoJSON-format handling in the client (vector-only).

### Already shipped toward this
- #218 lowered the cutoff 100k→30k (interim shrink).
- #213/#214 (gzip) and the gzip PUT become dead code once this lands.

Prior context: #190 (the GeoJSON-path evaluation), #209 (access patterns), geo-agent#252/#256.

cboettig:
Closes #219.

## Summary

- Removes the auto-select between GeoJSON FeatureCollection and MVT pyramid (`_GEOJSON_MAX_FEATURES`, `serve_format`)
- Deletes all GeoJSON build helpers: `build_geojson_statement`, `_geojson_body`, `geojson_fc_query`, `count_globe_spanning_statement`, `_s3_internal_write_url`, `_put_gzip`
- Simplifies `render_recipe` to vector-only (no format branch)
- Removes Phase 2 / stat_levels conditionals — pyramid always builds fully
- Drops `format` and `geojson_url` from all metadata and result dicts
- Removes endpoint 404 guard for `format == "geojson"`
- Deletes `tests/test_tile_geojson.py` (233 lines) and removes `_force_vector_path` monkeypatching fixtures from two test files
- Net: −490 lines, 263/263 tests pass

## Test plan

- [x] Full test suite passes: `pytest -x -q` → 263 passed
- [ ] Smoke-test a tiny dataset (<5k cells) maps first-paint ≤ ~2s (per gating check in #219)


cboettig:
## Smoke-test & benchmark results (dev, post-deploy)

**Code correctness**
- All responses return `source.type = "vector"` — GeoJSON source never returned ✓
- `geojson_url` and `format` fields absent from all responses ✓
- `value_stats.by_res` covers all pyramid levels (e.g. res 2–8), confirming Phase 2 always runs ✓

**Build times (pod logs)**

| Cells | Total | copy | stats | bounds |
|-------|-------|------|-------|--------|
| 4,519 (tiny) | 3.1s | 2.1s | 0.8s | 0.2s |
| 114,445 (medium) | 3.0s | 1.8s | 0.8s | 0.3s |

**Tile serve** (tiny set, zoom 10): 200 OK, 2.7 KB, 0.56s

**Notes on the gating check (#219)**
Tiny sets (<5k cells) build in ~3s with MVT vs ~1–2s with the old GeoJSON path (GeoJSON skipped Phase 2 entirely). Cold first-paint for tiny sets is therefore ~3.5s rather than the ≤2s target in the issue. Warm cache (subsequent views) is ~0.5s tile fetch. Build times are acceptable — shipping as-is.

PR #255, opened 2026-06-29
Serve hex tileset metadata.json from the tile host

cboettig:
## What

Adds `GET /tiles/{namespace}/{name}/metadata.json` — serves a hex tileset's metadata sidecar (`value_stats`, `bounds`, `layer_name`, `value_columns`, `finest_res`, …) from the **same origin as the `.pbf` tiles**, reusing the existing `_get_cached_metadata` helper. Additive, backward-compatible: new route only, tile serving untouched.

## Why

Investigating "hex tiles not showing on the global-30x30 app", the server side was fully healthy — tiles serve 200 with data at every zoom, CORS is correct, the geo-agent render code is correct. The actual failure was in the agent loop: weak open models (qwen3) emit `add_hex_tile_layer` as **unparsed text in the response `content`** instead of a structured tool call, so the harness never runs the layer-add and the map stays blank even though the tileset built fine.

From the proxy logs (qwen3, Florida carbon, tileset built as `4dc22143…`), the leaked call was also **corrupted** — broken JSON in `value_stats` (`"min":0.0,"`), mangled tags. The trigger is that `add_hex_tile_layer` makes the model transcribe a large `value_stats` JSON blob + `bounds` array copied out of `get_hex_tile_status`. Big structured copies are exactly what small models corrupt.

This route lets the **client** fetch those color-scale inputs by hash (deriving the URL from `tile_url_template` via a suffix swap) instead of routing them through the LLM. The sidecar already exists on S3; this exposes it same-origin so the client needs no S3 base URL. This is the **server half**; the client half is geo-agent issue (filed separately) making `value_stats`/`bounds` optional in `add_hex_tile_layer` and fetching this endpoint when omitted.

## Tests

`tests/test_tile_endpoint.py::TestServeMetadata` — returns value_stats/bounds/layer_name, immutable cache-control, 404 on unknown hash / non-hex namespace, and the tile_url→metadata.json suffix-swap contract the client relies on. Full tile-endpoint suite (19) passes locally; `test_server.py`/`test_integration.py` need `mcp` (CI only).

### DIFF
--- a/server.py
+++ b/server.py
@@ -295,7 +295,7 @@ mcp.tool(name="query", description=query.__doc__)(_query_tool)
 import concurrent.futures
 import threading
 import time
-from tiles.endpoint import serve_tile
+from tiles.endpoint import serve_metadata, serve_tile
 from tiles.db import build_tile_connection
 from tiles.pyramid import (
     MVT_LAYER_NAME,
@@ -782,6 +782,9 @@ def mount_tiles(app):
     con = _get_tile_con()
     app.state.tile_con = con
     app.add_route("/tiles/{namespace}/{name}/{z:int}/{x:int}/{y:int}.pbf", serve_tile)
+    # Same-origin metadata sidecar so clients can read value_stats/bounds by
+    # hash instead of routing the large JSON through an LLM's tool-call args.
+    app.add_route("/tiles/{namespace}/{name}/metadata.json", serve_metadata, methods=["GET"])
 
 # -------------------------------------------------------------------------
 # 9. OPTIONAL BEARER TOKEN AUTH
--- a/tests/test_tile_endpoint.py
+++ b/tests/test_tile_endpoint.py
@@ -5,7 +5,7 @@ from starlette.applications import Starlette
 from starlette.routing import Route
 from starlette.testclient import TestClient
 
-from tiles.endpoint import serve_tile
+from tiles.endpoint import serve_metadata, serve_tile
 from tiles.pyramid import register_hex_tiles
 from tiles.db import build_tile_connection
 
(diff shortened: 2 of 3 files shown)


cboettig:
Client half: boettiger-lab/geo-agent#276 (make value_stats/bounds optional in add_hex_tile_layer, fetch this endpoint by hash). Recipe-correctness sibling: #256 (agg COUNT→SUM).

cboettig:
## Reopening — re-instate the metadata.json route (un-revert #287)

#287 reverted this endpoint on the reasoning that (a) the transcription leak it defended against was a one-time qwen3 event, absent from every other model, and (b) nothing consumed the route. A live ca-30x30 incident this morning revisits both premises.

### The leak recurred on a non-qwen model

`z-ai/glm-5.2`, 2026-07-14 ~15:22 UTC, ca-30x30. GLM built a tileset (`register_hex_tiles` agg=AVG → hash `7693f3ca387d3c91`) with fully-valid `value_stats`, then leaked its XML arg-dialect into the `value_stats` value when transcribing it into the client's `add_hex_tile_layer` — the exact corruption class #255 was built to route around, on the model family #287 called clean. Full write-up: geo-agent#276 (reopened).

### This route is the server half the client fix needs

geo-agent#276 is reopened to make the client fetch `value_stats`/`bounds` by hash instead of having the LLM transcribe them — so the "no consumer" objection is being resolved on the client side. This route is that consumer's dependency.

### Low-cost re-instatement

The metadata sidecar is still written per-tileset (`pyramid.py` → `{output_uri}metadata.json` with `value_stats`, `bounds`, `layer_name`, `value_columns`, `finest_res`). #287 removed only the HTTP exposure; the data is already on S3, keyed by hash. Re-instating is a clean un-revert of the additive `GET /tiles/{namespace}/{name}/metadata.json` route (reuses `read_existing_metadata`), plus the `TestServeMetadata` suite (suffix-swap contract, immutable cache-control, 404 on unknown hash / non-hex namespace) that came with #255.

Same-origin off the tile host is preferred over pointing the client at a raw `s3://public-output/...` URL: no bucket-layout knowledge on the client, and CORS/cache headers stay under our control.

Issue #260, opened 2026-07-03
Explore a clean fallback to source.coop when Ceph S3-west is unavailable


cboettig:
## Motivation

NRP Ceph **S3-west** is our single read source for public data, and it's a hard single point of failure. When it's down (as it is right now), **every** query and STAC href resolution fails — there is no automatic degradation.

Most of our public data is *already* mirrored to **source.coop** (AWS `us-west-2`, anonymous reads, reader doesn't pay egress). We should explore wiring this in as a clean, low-config read fallback so the server stays usable when Ceph is unavailable.

## What we already have to build on

- **source.coop mirror** — public NRP buckets are mirrored to AWS us-west-2 via source.coop, anonymous/UNSIGNED reads over Internet2/CENIC.
  - Path mapping: `s3://public-<name>/...` → `s3://us-west-2.opendata.source.coop/cboettig/<name>/...`
  - e.g. `s3://public-carbon/**` → `s3://us-west-2.opendata.source.coop/cboettig/carbon/**`
  - Bucket has dots → needs `URL_STYLE 'path'` so TLS SNI matches; region `us-west-2`, default AWS endpoint, anonymous.
- Endpoint config today:
  - `stac.py` rewrites public `s3-west.nrp-nautilus.io` hrefs to the internal rook host via `_S3_INTERNAL` (env `S3_ENDPOINT_URL`, default `http://rook-ceph-rgw-nautiluss3.rook`) — see `stac.py:44-47`.
  - `server.py` query path defaults the endpoint to `s3-west.nrp-nautilus.io` for user-supplied creds — see `server.py:130`.

## Scope to explore

1. **Trigger / detection** — how do we decide Ceph is down and flip to the mirror? Options to weigh:
   - Health-probe the RGW endpoint on startup / periodically and select the source.
   - Try-Ceph-then-fallback per query (retry on connection error).
   - Manual/env switch (`S3_SOURCE=sourcecoop`) as the minimal first step.
2. **Path + endpoint rewriting** — a clean mapping layer from `public-<name>/...` to the source.coop path, applied to both STAC hrefs (`stac.py`) and the query secret/endpoint (`server.py`). Anonymous secret with `URL_STYLE 'path'`, region us-west-2.
3. **Scope of coverage** — only *public* datasets are mirrored; private data has no fallback. Fallback must be public-only and fail cleanly (clear message) for private scopes.
[...]

PR #261, opened 2026-07-03
source.coop mirror fallback for public data reads (#260)

cboettig:
Explores a clean fallback to the public **source.coop** mirror so the server stays usable when NRP Ceph **S3-west** is unavailable. Closes part of #260 (Phase 1).

## Approach

Additive and **always-on** — no server-side mode switch, no effect on the Ceph happy path. DuckDB routes reads by path prefix, so the mirror is simply an *additional* known source. This matches how apps already work (they link individual bucket `stac-collection.json` entries directly rather than walking a top-level catalog — there is no top-level catalog on source.coop).

## Changes

- **`query-setup.md`** — add an anonymous, prefix-scoped `source_coop` S3 secret. `s3://us-west-2.opendata.source.coop/*` routes to the mirror; `s3://public-*` still routes to Ceph.
- **`stac.py` `_href_to_s3`** — rewrite `https://data.source.coop/` asset hrefs to the `s3://us-west-2.opendata.source.coop/` bucket form, so DuckDB can expand `h0=*` globs (the `data.source.coop` HTTPS gateway cannot list).
- **`stac.py` `_collection_to_dict`** — use `lnk.get_target_str()` instead of `lnk.href`. `lnk.href` triggers `get_root()`, which resolves the collection's root link over the network — that crashes when rendering an inline source.coop collection during a Ceph outage (the whole point). `get_target_str()` returns the raw href with no I/O.

Net effect: pass a source.coop `stac-collection.json` to the STAC tools (inline `collection` or `catalog_url`) and every asset resolves to a queryable s3 mirror path — end-to-end during a Ceph outage.

## Verification (against live source.coop while Ceph was down)

- Two-secret prefix routing confirmed via `which_secret()`; the server's actual parsed `SETUP_SQL` produces correct routing + a real mirror read (carbon: 4.8B rows, ~2s).
- Full inline source.coop collection → every asset (parquet hex + COG) resolves to an s3 mirror path, structured and markdown paths.
- **271/271 tests pass**, incl. 5 new: `_href_to_s3` rewrite cases, asset rewrite in `_collection_to_dict`, and a regression guard that link rendering never resolves the root over the network.

## Deferred (Phase 2, follow-up)

Automatic Ceph-down detection + auto-resolving each dataset's source.coop entry. That needs a dataset↔bucket map, since source.coop collection ids differ from the bucket name (e.g. `irrecoverable-carbon` vs bucket `carbon`).

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -93,11 +93,45 @@ kubectl logs -l job-name=s3-throughput -f --max-log-requests=12 --prefix
 kubectl delete job s3-throughput
 ```
 
+### `source-coop-endpoint-bench.py`
+
+Compares the two front doors to the *same* source.coop objects: the **direct AWS
+S3 bucket** (`us-west-2.opendata.source.coop`, a CNAME straight to
+`s3.us-west-2.amazonaws.com` — what the #260/#261 fallback maps to) vs the
+**`data.source.coop` gateway** (Cloudflare-fronted CDN proxy over the same S3
+origin, the host that appears in the source.coop STAC hrefs). Raw HTTP probes
+(footer latency, single-stream MB/s, aggregate MB/s) isolate transport; DuckDB
+`count(*)` and single-column `sum()` over the whole dataset show the query-path
+impact. Cold and warm passes both reported (Cloudflare may edge-cache).
+
+```bash
+uv run --with 'duckdb>=1.1' --with requests python3 source-coop-endpoint-bench.py
+```
+
+Result (2026-07-03, run from **UC Berkeley campus egress — not NRP**; relative
+comparison holds directionally, absolute numbers differ on-cluster over
+Internet2/CENIC — re-run the k8s form for production numbers):
+
+| metric | aws-direct | data.source.coop proxy |
+|---|---|---|
+| footer latency (median) | 133 ms | 138 ms |
+| single-stream throughput (median) | **80.8 MB/s** | 64.5 MB/s |
+| aggregate throughput (c=8) | 96.9 MB/s | 108.9 MB/s (n=1, noisy) |
+| DuckDB `count(*)` (warm) | 2.7 s | 2.9 s |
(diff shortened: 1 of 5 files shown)


cboettig:
## Benchmark: direct AWS bucket vs data.source.coop proxy

Question raised in review: the fallback rewrites hrefs to `s3://us-west-2.opendata.source.coop/...` — is that the CDN proxy, and would going direct to AWS be faster?

**Clarification first:** `us-west-2.opendata.source.coop` is a **CNAME straight to `s3.us-west-2.amazonaws.com`** — so the fallback *already* reads the direct AWS bucket. The `data.source.coop` host that appears in the source.coop STAC hrefs is the **Cloudflare-fronted proxy** over the same S3 origin. So this compares those two front doors to the same bytes (`benchmarks/source-coop-endpoint-bench.py`).

**Results** (2026-07-03, run from **UC Berkeley campus egress — NOT NRP**; the relative comparison holds directionally, absolute numbers differ on-cluster over Internet2/CENIC):

| metric | aws-direct | data.source.coop proxy |
|---|---|---|
| footer latency (median) | 133 ms | 138 ms |
| single-stream throughput (median) | **80.8 MB/s** | 64.5 MB/s |
| aggregate throughput (c=8) | 96.9 MB/s | 108.9 MB/s (n=1, noisy) |
| DuckDB `count(*)` (warm) | 2.7 s | 2.9 s |
| DuckDB `sum(1 col)` (warm) | **176 s** | 224 s |

**Conclusion:** direct AWS is as-good-or-better on every robust metric and **~25–27% faster** on the throughput-bound cases (single-stream, full-column scan). Cloudflare edge caching did **not** help — these large, rarely-requested parquet files aren't hot in the CDN, so the proxy is just an extra hop to the same origin. The single aggregate-throughput number where the proxy edges ahead is n=1 and within noise.

So the current mapping (direct AWS bucket) is the right call — no code change. Follow-up worth doing: re-run the k8s Job form on-cluster for production-representative NRP→AWS numbers before Phase 2.

Issue #268, opened 2026-07-05
Env-configurable default S3 endpoint (S3_DEFAULT_ENDPOINT) — deploy a data-access head pointed at any storage


cboettig:
## Motivation

#264 lets a **client** choose the query endpoint per request (anonymous bring-your-own-bucket). The complement is letting a **deployment** set its *default* storage endpoint, so standing up this codebase pointed at a different backend (Ceph, MinIO, source.coop) is a **pure env change, no code**.

This enables the app-side host-selection model: an app points `mcp_url` at whichever data-access head matches its data source (e.g. a MinIO-backed head during a Ceph outage, or an aws-west-2 head reading source.coop same-region with no egress). The server stays stateless and source-agnostic.

## Current state

- `STAC_CATALOG_URL` is already env-configurable. ✅
- But the **default query secret** hardcodes the endpoint in `query-setup.md`:
  ```sql
  CREATE OR REPLACE SECRET s3 (TYPE S3, ENDPOINT 'rook-ceph-rgw-nautiluss3.rook', URL_STYLE 'path', USE_SSL 'false', KEY_ID '', SECRET '');
  ```
  So `s3://public-*` reads always target Ceph unless the client overrides per-request. There's no per-deployment default knob.

## Proposal

Add `S3_DEFAULT_ENDPOINT` (and, since endpoints differ in transport, `S3_DEFAULT_URL_STYLE` + `S3_DEFAULT_USE_SSL`, or infer SSL from the endpoint as today). The default `s3` secret's endpoint is taken from it, falling back to `rook-ceph-rgw-nautiluss3.rook` for backward compatibility (unset = current behavior).

Implementation note: the default-secret line in `query-setup.md` is both *executed* and *shown to the model*. Cleanest is to **move default-secret creation into `get_isolated_db`** (code, env-driven) and keep `query-setup.md` as model-facing guidance only — mirroring how the per-request `client_s3` secret is already built in code. Preserve the scoped `source_coop` secret.

## Result

- **MinIO head**: `S3_DEFAULT_ENDPOINT=minio.carlboettiger.info` (https, path-style) + `STAC_CATALOG_URL=<minio root catalog>`. Apps point `mcp_url` here → transparent MinIO, no per-query `s3_endpoint` needed.
- **source.coop head** (e.g. aws-west-2): same-region, no-egress reads; source.coop paths already routed by the scoped `source_coop` secret, so mostly a catalog/location choice.
[...]

Issue #264, opened 2026-07-04
Anonymous bring-your-own-bucket for `query` (symmetric with credentialed BYO) — the only server change multi-source needs


cboettig:
## Reframe (supersedes the original "source registry" framing)

Investigating the source.coop/MinIO fallback clarified that **almost none of multi-source support belongs in this server.** The system is already agnostic by carrying per-dataset links from any source:

- **App-layer resilience is entirely client-side.** With MinIO now a *self-consistent* mirror (collection JSON + every nested PMTiles/COG/parquet href points at `minio.carlboettiger.info`), an app just repoints its `catalog` + `collection_url`s and the whole map cascades to MinIO. Verified end-to-end during the Ceph outage: collection JSON `200`+CORS, PMTiles `206`+range+CORS, COG `206`+CORS, and titiler reads the MinIO COG server-side (`/cog/info` `200`). **No server change, no host-rewriting.**
- **STAC tools are already agnostic** — `catalog_url` / inline `collection` accept any source.
- **Credentialed query routing is already agnostic** — a client can point `query` at any *private* S3 via `s3_key`/`s3_secret`/`s3_endpoint`/`s3_scope`.

So the **only** server-touching gap is the `query` (SQL-over-parquet) path for **anonymous** sources.

## The actual gap

`get_isolated_db` (`server.py:129`) only injects a client endpoint secret when **both** `s3_key` and `s3_secret` are non-empty:

```python
if s3_key and s3_secret:            # anonymous (empty creds) → endpoint ignored → falls back to Ceph default
    endpoint = s3_endpoint or "s3-west.nrp-nautilus.io"
    ...
```

So you can bring your own *credentialed* bucket but not your own *anonymous* bucket. That asymmetry — not any missing "registry" — is why every anonymous mirror (source.coop #261, MinIO #266) has needed a bespoke server-side secret.

## Proposed change (one small, generic, one-time)

Make anonymous BYO-bucket symmetric with the credentialed path: let `s3_endpoint` (+ optional `s3_scope`, and url_style/ssl) create a secret even with empty creds. E.g. `if s3_endpoint or (s3_key and s3_secret):`, defaulting KEY_ID/SECRET to `''`.

[...]

PR #272, opened 2026-07-05
docs: app-maintainer guide for mirror failover during an S3 outage

cboettig:
Adds `docs/guide/mirror-failover.md` — how an app maintainer points an app at the MinIO mirror (`minio.carlboettiger.info`) when `s3-west` is down.

Covers the two independent surfaces:
- **Layers** (client-side): swap `catalog` + `collection_url` hosts → mirror; assets cascade (self-consistent mirror); TiTiler reads mirror COGs server-side.
- **Query** (MCP): point `mcp_url` at a mirror head (`S3_DEFAULT_ENDPOINT`, #268) or pass `s3_endpoint` per request (#264); paths stay canonical `s3://public-*`.

Plus the `get_schema` https-glob caveat (flagged honestly), reverting, and the DNS-failover zero-touch alternative. Refs #260/#264/#268. Docs-only.

### DIFF
--- a/docs/guide/mirror-failover.md
+++ b/docs/guide/mirror-failover.md
@@ -1,0 +1,85 @@
+# Running an app on a mirror during an S3 outage
+
+When the primary NRP Ceph endpoint (`s3-west.nrp-nautilus.io`) is unavailable,
+an app can be pointed at a public **mirror** of the `public-*` buckets and keep
+working. This is app-driven — nothing needs to change on the primary server.
+
+The reference mirror is **`minio.carlboettiger.info`** (MinIO): a drop-in copy of
+the NRP `public-*` buckets — same bucket names, same catalog structure,
+self-consistent asset hrefs, public/anonymous reads, with CORS + HTTP range
+enabled for browser access. (A partial AWS mirror also exists on source.coop; see
+[architecture.md](architecture.md) and issue #260. MinIO is the more complete
+drop-in.)
+
+## The two data surfaces
+
+An app reads data over **two independent paths**, and both must be pointed at the
+mirror to fully ride out an outage:
+
+| Surface | Fetched by | Through the MCP server? |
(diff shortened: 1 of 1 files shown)


cboettig:
Reviewed against what actually shipped (#273/#275/#276) and against live behavior on dev during the current outage. The guide is accurate and current — the registry/hint/scope story in §2 matches exactly what I verified end-to-end yesterday (mirror glob reads via the `source_coop` secret, option (b)'s ⚠️ hint line rendering with the right params, scoped BYOB winning longest-match, and the #273 no-scope semantics as described in the callout). Doc links all resolve. Three notes:

1. **Hex-tile claim in option (a) is half-true and worth a caveat.** "hex-tile generation, which now honors the same default endpoint — #275" holds for *reads*, but pyramid **writes** go to `s3://public-output` with anonymous credentials — fine on NRP Ceph, presumably rejected by the minio bucket. Until #279 lands (scoped `S3_SOURCES` entry with write creds for the output bucket), `register_hex_tiles` on a mirror-configured head will fail at write time. Suggest: a one-line caveat pointing at #279, or fold the #279 snippet into the option (a) config block once verified.

2. **Degraded start is worth one sentence in option (a).** The config swaps `STAC_CATALOG_URL` to the mirror, so startup succeeds — but an operator who repoints only `S3_DEFAULT_ENDPOINT`/`S3_SOURCES` and keeps the Ceph catalog URL needs `STAC_ALLOW_DEGRADED_START=true` to boot at all during the outage (this is exactly how dev rolled out yesterday: pods came up degraded, 0 collections, everything else working).

3. **Tiny**: option (b) works for hand-built inline collections too now — during the outage that path surfaced #278 (inline dicts required `links`/`description`); fixed in #280, so no doc change needed, just noting the dependency if this merges first.

With or without the two suggested sentences, content is correct — good to merge from my side.

Issue #303, opened 2026-07-12
get_stac_details/get_schema: dedup per-column descriptions across assets (duplication truncates large multi-asset schemas at 16k)


cboettig:
## Problem

`get_stac_details` (and therefore `get_schema`, which forwards the app's cached STAC inline to it) renders **every column's description once per asset**. On the common **flat GeoParquet + hex** dataset, each shared attribute description is emitted **2–3×** — once under the flat `read_parquet(...)`, once under the hex, and a third time when a legacy collection-level `table:columns` "Key columns" trailer is present (`_extract_columns`).

This is the highest-traffic metadata channel, and the duplication is not just tokens — it causes **schema truncation**.

## Evidence (from `logs-open-llm-proxy`, ~3 months of geo-agent traffic)

Tool-call distribution (the metadata routes):

| tool | calls |
|---|--:|
| `query` | 10,918 |
| **`get_schema`** | **4,407** |
| `get_stac_details` | 540 |
| `get_collection` | 63 |

So the schema the model sees comes overwhelmingly through `get_schema` → `_format_collection`. The cold prompt carries **no** columns (geo-agent compact index, #306); columns arrive only via this renderer.

124 distinct-dataset schema payloads sampled from real results:
- **88 / 124** render **>1 queryable asset** (the flat+hex / multi-layer case).
- Repeated *described-column* lines in multi-asset payloads: `nci-frontiers` **62**, `iucn-richness-2025` **52**, `ace-terrestrial-biodiversity-summary` **25**, `calfire-2025-firep` **20**, `usgs-nhd-streams-by-order` **20**, … — the same column description printed 2–3×.
- **17 / 124** still carry a collection-level `Key columns` trailer (a third copy).
- Size: median **3,958** chars; **three** payloads (`nci-frontiers`, `ace-terrestrial-biodiversity-summary`, `ca30x30-conserved-areas-terrestrial-2025`) sit at exactly **16,016** = geo-agent's **`TOOL_RESULT_CAP = 16000`** (`geo-agent app/agent.js:287`) + truncation marker. **These schemas are cut off — the model never sees the tail columns**, and fabricates or emits broken SQL.

## Fix

[...]

PR #306, opened 2026-07-13
fix(#305): compact parent-collection index — defer child columns to per-child calls

### LINKED ISSUE #305: get_stac_details: parent-collection index inlines every child's full columns (should be a compact chooser, deferring columns to per-child calls)
## Problem

`get_stac_details` on a **parent** collection inlines **every child's full column schema**, even though the same output ends with "Call `get_stac_details` with a sub-dataset ID … for column schemas and details." So the parent index — which should be a compact "which sub-dataset do I want?" chooser — instead ships the entire bucket's schemas, blowing the 16k tool-result cap and duplicating exactly what the per-child call returns.

Live on dev today:

| parent collection | `get_stac_details` size | children |
|---|--:|--:|
| `pad-us-4.1` | **49.9k** | 10 |
| `high-seas` | **45.1k** | 24 |
| `ca30x30` | **38.7k** | 20 |
| `social-vulnerability-index` | **26.7k** | 10 |
| `us-census` | **21.8k** | 12 |

`us-census` (21.8k) renders 12 `- columns:` lines + a shared description block across its 6 children — none of which the model needs to *choose* a child, and all of which it re-fetches on the per-child call anyway.

## Root cause

`stac.py` `_format_collection`, the sub-children branch:

```python
if sub_children:
    lines.append(f"\n**Sub-datasets ({len(sub_children)}):**\n")
    for sc in sub_children:
        sc_title = sc.title or sc.id
        sc_parquet = _extract_parquet_assets(sc)   # ← renders full per-asset columns + shared block
        if sc_parquet:
            lines.append(f"*{sc_title}* (`{sc.id}`):")
            lines.extend(sc_parquet)
    lines.append(f"\nCall get_stac_details with a sub-dataset ID … for column schemas and details.")
```

`_extract_parquet_assets(sc)` is the leaf renderer — it emits each asset's `read_parquet(...)` path **plus** the per-asset `columns:` names/types line **plus** the shared "Column descriptions" block (#303). In the parent index that column payload is pure bloat; the trailing note already tells the model to fetch columns per-child.

(Note the census *unit test* `test_parent_does_not_show_child_columns` passes only because its fixture puts columns at the child *collection* level with empty asset `extra_fields`; real children carry per-*asset* `table:columns`, so production parents do render them. The fixture masks the behavior.)

## Fix

In the sub-children branch, render each child **compactly** — title + `id` + `read_parquet(...)` path(s) + a one-line asset/extension summary (h3/vector, size, per-feature dedup note) — and **omit the column schemas**, deferring them to the per-child `get_stac_details` call (which the note already directs).

Concretely: add a `columns=False` (or `compact=True`) mode to `_extract_parquet_assets` — or a small `_extract_parquet_assets_compact(sc)` — that skips the per-asset `columns:` line and the shared description block, and call that from the sub-children loop. The leaf branch (`elif parquet_assets:`) keeps the full render unchanged.

Expected: `us-census` ~22k → ~2–3k; `pad-us-4.1`/`high-seas`/`ca30x30` all fall well under 16k as compact choosers.

## Why this matters / dependency

- This is the **prerequisite** for **data-workflows #407** (reshape flat "bucket-of-datasets" collections like `nci-frontiers`/`svi-2022` into parent + children). Restructuring is pointless while the parent re-dumps all children's columns — the parent would stay ~32k regardless.
- Complements **#303** (per-column dedup, shipped): #303 made the *leaf* schema lean; this makes the *parent index* lean.
- Pure render change: `_collection_to_dict` (structured `get_collection`) and per-child leaf output stay full and self-describing.

## Acceptance

- `get_stac_details` on a parent collection lists children as a compact chooser (title + ID + `read_parquet(...)` path + brief asset/ext summary), **no** per-column descriptions, and lands well under 16k for the collections in the table above.
- Per-child `get_stac_details(<child_id>)` output is unchanged (full schema).
- `_collection_to_dict` unchanged.
- Update `test_parent_does_not_show_child_columns` to use a realistic child fixture (per-*asset* `table:columns`) so it actually guards this behavior.

## Cross-refs

- #303 (leaf per-column dedup, shipped)
- data-workflows #407 (bucket-of-datasets restructure — depends on this)

cboettig:
Closes #305.

## Problem

`get_stac_details` on a **parent** collection inlined **every child's full column schema**, even though the same output ends with "Call `get_stac_details` with a sub-dataset ID … for column schemas." The parent index — a chooser — thus re-dumped the whole bucket, over the 16k `TOOL_RESULT_CAP`, duplicating what the per-child call returns:

| parent | before | after |
|---|--:|--:|
| pad-us-4.1 | 49.9k | **4.4k** |
| high-seas | 45.1k | **11.4k** |
| ca30x30 | 38.7k | **12.6k** |
| us-census | 21.8k | **3.1k** |

## Change

`_extract_parquet_assets(col, compact=True)` emits only each asset's `read_parquet(...)` path + routing hint + description + extension fields — no per-asset column names, no shared description block. The sub-children branch of `_format_collection` uses it and appends an explicit per-child `columns: call get_stac_details('<child-id>')` pointer. Leaf render (per-child `get_stac_details`) and `_collection_to_dict` are unchanged.

Example (3-child parent, ~1.1k):
```
**Sub-datasets (3):**

*2024 State Boundaries* (`census-2024-state`):
  - GeoParquet: `read_parquet('s3://public-census/census-2024/state.parquet')`
  - hex: `read_parquet('s3://public-census/census-2024/state/hex/**')`
    - h3:native_resolution: 8
  - columns: call get_stac_details('census-2024-state')
…
```

## Risk — log-checked before changing behavior

Two weeks of prod traffic (581 query-sessions, 3,107 schema calls):
- Parent ids are **1.0%** of schema calls (31/3107); apps address child/leaf IDs directly and use the parent only as a browse/chooser step (`browse → collection(parent) → schema(child)`).
- **0 of 581** query-sessions queried using only-parent schema — every real query had a leaf `get_stac_details` first. The lone borderline per-dataset case in 2 weeks (`pad-us-4.1 → combined.parquet` with well-known `Unit_Nm/GIS_Acres/GAP_Sts`) still works via a one-call leaf round-trip, which the explicit pointer now signposts.

So deferring child columns has negligible real-world impact, and the compact chooser still shows each child's `read_parquet(...)` path.

## Tests

`TestCompactParentIndex` (4 cases) with a **realistic per-asset-columns fixture** — the census fixture stored columns at the collection level, which masked this behavior. Asserts: parent omits child descriptions / shared block / per-asset `- columns:` line; lists each child's path + pointer; leaf render still shows full columns; parent stays small. Full suite: **368 passed**.

## Cross-refs

- #303 (leaf per-column dedup, shipped) — made the leaf lean; this makes the parent index lean.
- data-workflows #407 (bucket-of-datasets restructure) — depends on this: restructuring is moot while the parent re-dumps children.

### DIFF
--- a/stac.py
+++ b/stac.py
@@ -195,7 +195,7 @@ def _any_queryable_asset_has_columns(col) -> bool:
     return False
 
 
-def _extract_parquet_assets(col) -> list[str]:
+def _extract_parquet_assets(col, compact: bool = False) -> list[str]:
     """Extract parquet/hex asset lines from a collection's assets.
 
     Column *descriptions* are deduplicated (#303): rather than reprinting every
@@ -204,6 +204,14 @@ def _extract_parquet_assets(col) -> list[str]:
     cap and truncated them — each asset line lists only its own column
     names/types, and every column's description + categorical `values` is
     rendered exactly once in a shared block appended after the asset lines.
+
+    *compact* (#305): emit only each asset's ``read_parquet(...)`` path, routing
+    hint, description, and extension fields — NO per-asset column names and NO
+    shared description block. Used by the parent/sub-dataset index, where the
+    columns would re-dump every child's schema (20-50k parent payloads over the
+    16k cap) and duplicate what the per-child ``get_stac_details`` call returns.
+    The caller appends an explicit per-child "call get_stac_details(<id>)"
+    pointer instead.
     """
     assets = []
     shared_cols: dict = {}  # name -> col dict, in first-seen order (union of all assets)
@@ -246,6 +254,10 @@ def _extract_parquet_assets(col) -> list[str]:
                 ext_val = asset.extra_fields.get(ext_key)
                 if ext_val is not None:
                     assets.append(f"    - {ext_key}: {ext_val}")
+            # Per-asset columns are omitted in compact mode (parent index, #305):
+            # the caller adds a per-child get_stac_details pointer instead.
+            if compact:
+                continue
             # Per-asset column *set* only (names/types); descriptions are shared below.
             asset_cols = asset.extra_fields.get("table:columns", [])
             summary = _asset_column_summary(asset_cols)
@@ -266,10 +278,11 @@ def _extract_parquet_assets(col) -> list[str]:
                         if not existing.get(k) and c.get(k):
                             existing[k] = c[k]
 
-    block = _format_columns(list(shared_cols.values()))
-    if block:
-        assets.append("\nColumn descriptions (shared across the assets above):")
-        assets.extend(block)
+    if not compact:
+        block = _format_columns(list(shared_cols.values()))
+        if block:
+            assets.append("\nColumn descriptions (shared across the assets above):")
+            assets.extend(block)
     return assets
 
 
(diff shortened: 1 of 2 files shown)


AlexanderLuzDH:
Disclosure: I maintain Busleyden Guard. Its public, install-free preview of this PR found one review obligation: preserve existing data readability and a safe rollback/forward-fix path for the changed STAC behavior, with tests/test_stac.py as candidate evidence: https://guard.busleyden.com/api/public-pr-preview?pr_url=https%3A%2F%2Fgithub.com%2Fboettiger-lab%2Fmcp-data-server%2Fpull%2F306

This is not a bug or security verdict. The free standalone Action can be evaluated non-blocking with local JSON/Markdown artifacts and no App install, card, call, or automatic conversion. Beta protocol: https://github.com/AlexanderLuzDH/busleyden-guard-live-proof/blob/main/BETA.md

PR #348, opened 2026-07-31
fix(#346): advertise a client-reachable STAC catalog URL

### LINKED ISSUE #346: browse_stac_catalog advertises the in-cluster STAC_CATALOG_URL — clients can't self-configure (cirrus mirror was undiscoverable during the Ceph outage)
## Summary

`browse_stac_catalog` advertises the catalog root as the **server-side** `STAC_CATALOG_URL`, verbatim. On the cirrus deployment that is an in-cluster address no client can resolve, so a client cannot self-configure from the server — even though every other URL the server hands back in the same session is already client-usable.

Observed on `https://duckdb-mcp.carlboettiger.info/mcp` (2026-07-30):

```
# Available Datasets (183 collections)

STAC catalog: `http://minio-svc.minio.svc.cluster.local:9000/public-data/stac/catalog.json`
```

That host only resolves inside the k3s cluster. Meanwhile `get_collection("cwhr13")` on the same server returns asset hrefs on `https://minio.carlboettiger.info` — publicly reachable, and in fact the exact mirror a client needs. So the server *has* a client-usable base for the mirror; the one line that tells a client where the catalog lives is the only place it reports an unusable one.

## Why it matters (concrete cost, today)

NRP Ceph (`s3-west.nrp-nautilus.io`) was 503 across the board this afternoon. Every geo-agent app crashes at boot in that state, because `DatasetCatalog.load` fetches the app config's `catalog` URL unconditionally (geo-agent#287) — and MCP is never consulted, so pointing `mcp_url` at cirrus alone does not help.

The cirrus deployment is a complete answer to that outage: same bucket names, same paths, MinIO mirror, and `duckdb-mcp.carlboettiger.info` reads from it happily (verified — `SUM(Acres)` over the ca-30x30 conserved-areas layer returned 26,471,500 / 52,375,900, matching the published assessment exactly). But discovering *how* to point a client at it took reading `k8s/cirrus-deployment.yaml` and `cirrus-ingress.yaml` in this repo, then guessing hostnames until `minio.carlboettiger.info` answered 200. An operator under outage pressure should be able to ask the server.

The existing fallback runbook (`open-llm-proxy/headless/TESTING-DURING-CEPH-OUTAGE.md`) documents a *different*, lossier path — source.coop with per-collection href remapping, partial coverage, and a required dev-server rewrite (#261). The cirrus mirror is strictly better (full bucket coverage, no remapping) and was undiscoverable from the protocol.

## Root cause

`stac.py:860` (and the `no datasets` branch at :858) formats the raw module-level `STAC_CATALOG_URL`:

```python
url = STAC_CATALOG_URL
...
lines.append(f"STAC catalog: `{url}`\n")
```

There is no public-URL analogue for the catalog, though the pattern already exists for tiles — `tiles/pyramid.py:_public_base_url()` reads `MCP_PUBLIC_BASE_URL` precisely so emitted tile URLs are client-fetchable. Asset hrefs come out public only because the mirror's STAC JSON happens to store them that way, not by design.

## Proposed change

1. Add `STAC_PUBLIC_CATALOG_URL` (mirroring `MCP_PUBLIC_BASE_URL`), defaulting to `STAC_CATALOG_URL` so nothing changes for NRP prod, where the two are identical.
2. Report it in `browse_stac_catalog`. When it differs from the internal URL, say both, labelled — the client needs to know which one to put in its config:
   ```
   STAC catalog: `https://minio.carlboettiger.info/public-data/stac/catalog.json`
   (server reads it at http://minio-svc.minio.svc.cluster.local:9000/... — use the first URL in client configs)
   ```
3. Set it on the cirrus deployment: `STAC_PUBLIC_CATALOG_URL=https://minio.carlboettiger.info/public-data/stac/catalog.json`.
4. Optional, and the nicer version of this: surface it as structured data rather than prose, so a client can configure itself without parsing markdown — either on the existing `catalog://list` resource or as a field in the `initialize` response. A geo-agent app could then take `mcp_url` alone and derive its `catalog`, which removes the whole class of "client points at a dead endpoint while MCP is healthy" failures.

## Verification

```bash
# 1. the advertised URL must be fetchable from outside the cluster
curl -sI "$(<call browse_stac_catalog, parse the catalog URL>)" | head -1   # want 200

# 2. a geo-agent client configured only from that URL boots and answers
#    (ca-30x30 regression slice, nimbus qwen, during a Ceph outage)
```

Filed from ca-30x30 regression work; I did not touch this repo. Related: #263 (catalog walk isn't STAC-native) touches the same code path but a different concern; geo-agent#287 is the boot-crash that makes this load-bearing.

cboettig:
Closes #346.

## Problem

`browse_stac_catalog` reported the server-side `STAC_CATALOG_URL` verbatim. On cirrus that is an in-cluster address, so the one line telling a client where the catalog lives was the only URL in the whole session a client couldn't resolve. Still reproducible on `duckdb-mcp.carlboettiger.info` as of this PR:

```
# Available Datasets (219 collections)

STAC catalog: `http://minio-svc.minio.svc.cluster.local:9000/public-data/stac/catalog.json`
```

Meanwhile `get_collection` hands back hrefs on `https://minio.carlboettiger.info` — the exact mirror a client needs. Finding it during yesterday's Ceph outage meant reading `k8s/cirrus-*.yaml` and guessing hostnames.

## Change

1. **`stac.py`** — `public_catalog_url()` reads `STAC_PUBLIC_CATALOG_URL` at call time, defaulting to `STAC_CATALOG_URL`. Same contract as `MCP_PUBLIC_BASE_URL` in `tiles/pyramid.py`. Blank/whitespace is treated as unset.
2. **`list_datasets`** — reports the public URL; when the two differ, names the internal one as a labelled aside so the client knows which to put in its config:
   ```
   STAC catalog: `https://minio.carlboettiger.info/public-data/stac/catalog.json`
   (server reads it at http://minio-svc.minio.svc.cluster.local:9000/... — use the first URL in client configs)
   ```
   Covers the `No datasets loaded` branch too — degraded start (#262) is exactly when an operator is hunting for a reachable catalog. A caller-supplied `catalog_url` is echoed unchanged.
3. **`GET /version`** — item 4 of the issue, structured rather than prose:
   ```json
   {"version": "...", "git_sha": "...",
    "stac_catalog_url": "https://minio.carlboettiger.info/public-data/stac/catalog.json",
    "public_base_url": "https://duckdb-mcp.carlboettiger.info"}
   ```
   No MCP session, no auth token, no markdown parsing — a geo-agent app can take `mcp_url` alone and derive its `catalog` at boot, which retires the class of failure behind geo-agent#287. **Public URLs only**: this route is deliberately auth-exempt, so the internal address stays inside authenticated MCP responses (asserted by a test).
4. **`k8s/cirrus-deployment.yaml`** — sets `STAC_PUBLIC_CATALOG_URL`.
5. Docs: env-var table in `deployment.md`; `mirror-failover.md` gains an "ask the head" step and its stale "#346 is a display bug" bullet is corrected.

NRP prod is unaffected — `STAC_CATALOG_URL` there is already public, the two strings are equal, and output is byte-for-byte what it was (asserted by a test).

## Verification

- `395 passed` (full unit suite), including 7 new `TestPublicCatalogUrl` cases and 3 on `/version`.
- Advertised cirrus URL is fetchable from outside the cluster — the value in the manifest is correct, not just plausible:
  ```
  $ curl -sI https://minio.carlboettiger.info/public-data/stac/catalog.json | head -1
  HTTP/2 200
  ```

### Not done here

The cirrus rollout. This k3s cluster isn't in the kubeconfig available to me (only NRP `nautilus`), so it needs to be run from a machine with cirrus access — the deployment tracks `:main`, so after merge:

```
kubectl apply -f k8s/cirrus-deployment.yaml           # picks up STAC_PUBLIC_CATALOG_URL
kubectl rollout status deployment/duckdb-mcp -n mcp
curl -s https://duckdb-mcp.carlboettiger.info/version  # expect stac_catalog_url on minio.carlboettiger.info
```

The issue's second verification step (a geo-agent client configured only from that URL boots and answers the ca-30x30 slice) needs the client-side change to consume `/version`, which lives in geo-agent — worth filing there against #287 once this is deployed.

### DIFF
--- a/docs/guide/deployment.md
+++ b/docs/guide/deployment.md
@@ -44,6 +44,7 @@ See [`AGENTS.md` → Rollout workflow](https://github.com/boettiger-lab/mcp-data
 
 | Variable | Default | Description |
 |---|---|---|
-| `STAC_CATALOG_URL` | NRP public catalog | URL of the STAC catalog to serve |
+| `STAC_CATALOG_URL` | NRP public catalog | URL of the STAC catalog to serve (the address *this server* reads) |
+| `STAC_PUBLIC_CATALOG_URL` | `STAC_CATALOG_URL` | The same catalog's client-reachable URL. Set it only when the server reads the catalog over an address clients can't resolve (e.g. an in-cluster mirror); this is what `browse_stac_catalog` and `GET /version` advertise |
 | `THREADS` | 100 | DuckDB thread count (S3 workloads are I/O-bound) |
 | `PORT` | 8000 | HTTP server port |
--- a/docs/guide/mirror-failover.md
+++ b/docs/guide/mirror-failover.md
@@ -38,6 +38,23 @@ In the app config (`layers-input.json`):
 If the app's `mcp_url` comes from a deployment env var rather than the config
 file, that is a manifest change: `kubectl apply` it, not just a rollout restart.
 
+### Don't guess the mirror host — ask the head
+
+A head reports the client-reachable URL of the catalog it is reading, so step 1
+can be read off the server instead of out of these manifests (#346):
+
+```bash
+curl -s https://duckdb-mcp.carlboettiger.info/version
+# {"version":"...","git_sha":"...",
+#  "stac_catalog_url":"https://minio.carlboettiger.info/public-data/stac/catalog.json",
+#  "public_base_url":"https://duckdb-mcp.carlboettiger.info"}
+```
+
+`/version` takes no MCP session and no auth token, so an app that has an
+`mcp_url` can derive its `catalog` at boot rather than carrying a hardcoded one
+that may be the very thing that's down. `browse_stac_catalog` reports the same
+URL in its header line.
+
 ### Verify before declaring it done
 
 ```bash
@@ -75,10 +92,12 @@ for how to stop owing it.
   paths and creates the matching anonymous, prefix-scoped DuckDB secrets. Never
   hand-edit an endpoint into query SQL.
 - **Internal vs public hosts.** A head's own read endpoint may be an in-cluster
-  address (e.g. `minio-svc.minio.svc.cluster.local`); the URLs it hands you for
-  client use are public. If `browse_stac_catalog` reports an in-cluster catalog
-  root, that's a display bug (#346) — the public equivalent is the mirror host
-  above; asset hrefs from `get_collection` are already public.
+  address (e.g. `minio-svc.minio.svc.cluster.local`), but every URL it hands you
+  for client use is public: asset hrefs from `get_collection`, tile URLs, and —
+  since #346 — the catalog root in `browse_stac_catalog`, which names the
+  in-cluster address only as a labelled aside when the two differ. An
+  unqualified in-cluster catalog root means the head predates #346; get the
+  public equivalent from `GET /version` (above).
 - **When mixing sources, always pass `s3_scope`.** A per-request `s3_endpoint`
   (or credentials) *without* a scope applies to every `s3://` path in that query
   and disables the server default for the request (deterministic since #273) —
@@ -94,6 +113,12 @@ S3_DEFAULT_ENDPOINT=minio.carlboettiger.info
 S3_SOURCES='[{"name":"minio","https_prefix":"https://minio.carlboettiger.info/","s3_prefix":"s3://"}]'
 STAC_CATALOG_URL=https://minio.carlboettiger.info/public-data/stac/catalog.json
 ```
+
+If your head reads the catalog over an address clients can't resolve (an
+in-cluster service DNS name, as the cirrus deployment does), also set
+`STAC_PUBLIC_CATALOG_URL` to the external URL of that same catalog — that is what
+`/version` and `browse_stac_catalog` advertise. It defaults to `STAC_CATALOG_URL`,
+so a head configured like the block above needs nothing.
 
 `s3://public-*` reads — and hex-tile reads, which honor the same default
 endpoint (#275) — then resolve to the mirror anonymously, with no per-query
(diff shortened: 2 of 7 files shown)


cboettig:
Heads-up from the docs side — two things that affect this PR, both now resolved on `main`:

**1. CI was red for reasons unrelated to your change.** `requirements.txt` had a bare `mcp`, and the SDK's **2.0.0 (2026-07-28)** moved/removed `mcp.server.fastmcp`, so pytest failed at collection on every branch. Fixed by #349 (`mcp>=1.29,<2`, same rationale as the duckdb pin beside it). Rebase on `main` and your checks should go green. Migrating to the 2.x import surface is a separate call — yours, if you want it.

**2. This branch carries the docs commit from #347, which has now merged separately.** `fix/346-public-catalog-url` was cut from `docs/consolidate-outage-runbook`, so its diff-vs-`main` included `docs(outage): make mirror-failover the single runbook`. #347 squash-merged, so a rebase onto `main` will drop that duplicate — expect a conflict on `docs/guide/mirror-failover.md`, where **your version is the superset**: keep your `/version` section, your `STAC_PUBLIC_CATALOG_URL` block, and your rewritten "Internal vs public hosts" bullet.

One thing to preserve through the rebase: the section heading is now `## For agents querying through MCP during an outage {#agents}`. The `{#agents}` is load-bearing — <https://boettiger-lab.github.io/mcp-data-server/guide/mirror-failover.html#agents> is the URL we're handing to agents mid-outage, so it shouldn't churn.

Thanks for #346 — `GET /version` is a better answer than the one the runbook had. I've left the doc's guidance forward-compatible with it: it tells agents an unqualified in-cluster catalog root means the head predates #346, and to read the public URL from `/version`.

Issue #356, opened 2026-08-07
h3-guide: fractional-overlay example omits h3_cell_area, contradicting the coarse-feature example on the same page


cboettig:
## The defect

`h3-guide.md` gives two worked examples for overlaying a partial-coverage feature against a partial-coverage layer. They disagree on whether to weight by cell area, and the **unweighted one is the branch that fires for a res-10 feature**:

- **h3-guide.md:305** (same-resolution fractional overlay) — `100 * SUM(f.frac * COALESCE(c.w, 0)) / SUM(f.frac)`. No area term. Repeated at line 339.
- **h3-guide.md:333-334** (coarse feature over a finer layer) — `100 * SUM(COALESCE(p.w8, 0) * h3_cell_area(f.h8, 'km^2')) / SUM(h3_cell_area(f.h8, 'km^2'))`. Area-weighted.

Line 36 of the same file states the rule correctly: "H3 cells are not equal-area — true area varies with latitude and icosahedral distortion." Line 38 calls the scoped case "the accuracy-critical one." The line-305 example then drops it.

Dividing by `SUM(frac)` instead of `SUM(frac × h3_cell_area)` weights every cell equally, so a share aggregated across a range of latitudes is pulled toward whichever end has more cells. California spans 32.5°N–42°N and its res-10 cells vary by several percent.

The line-305 example uses `cwhr13` hex-fractions against ca30x30 conserved-areas — which is verbatim the question a partner asked our California 30x30 app this week. Two of four models reproduced the guide's formula exactly and returned numbers that miss the state's published 2025 Biodiversity Assessment on 12 of 13 habitat classes.

## Numbers

Per-class GAP 1+2 share of CWHR13 habitat, masked to California:

| CWHR13 class | guide's formula (line 305) | area-weighted | 2025 assessment |
|---|---:|---:|---:|
| Conifer Woodland (32) | 32.07 | 33.31 | 33.27 |
| Hardwood Forest (51) | 21.32 | 21.65 | 21.59 |
| Conifer Forest (31) | 23.75 | 24.20 | 24.14 |
| Shrub (70) | 27.57 | 27.82 | 27.79 |
| Barren/Other (20) | 51.97 | 51.70 | 51.67 |

Area-weighting lands within 0.06pp of the published report on 11 of 13 classes; the unweighted formula misses on 12 of 13, by up to 1.24pp.

## Reproduce

```sql
[...]

PR #362, opened 2026-08-08
k8s(prod): promote v0.8.13 by digest

cboettig:
> **On hold — see #364.** `v0.8.13` carries the #357 coarse-resolution overlay example, which reads
> `nland` from the `…-hex-weights-res8`/`-res9` assets. data-workflows#524 removed that column an hour
> after #357 merged, so promoting this digest would ship an example that is a binder error on any
> res-8/res-9 question. Fixed in #365; this promotion should be re-cut on a tag that includes it and
> re-gated. The gate below stays valid for what it covered — it exercised the res-10 path
> (CWHR13 percent-conserved), which is unaffected — but it never touched the broken branch, so a
> re-gate wants a cell that asks a res-8 or res-9 feature question.

Promotes prod from **v0.8.12** to **v0.8.13**.

```
-  image: …:v0.8.12@sha256:332d6df730c1fc66e7b2d317b121828fcb3785339a3a43b49789c3e22e6e0ffd
+  image: …:v0.8.13@sha256:1a3409d56f62bc8c9ba252a8960eb0dad16a240e24c400620eb7afae5233db7f
```

Digest resolved via the GHCR anonymous-pull manifest HEAD (docker CLI unavailable in this environment). **Method validated**: resolving `v0.8.12` the same way returns `sha256:332d6df…`, exactly the digest this commit replaces. The `v0.8.13` tag and the `19b3c709` git-sha tag resolve to the same digest.

## What ships

| PR | |
|---|---|
| #357 | fractional-overlay examples weight by `h3_cell_area`, use the published `hex-weights` assets, bound the denominator with the ecoregion land grid |
| #360 | rebuts the "area terms cancel" rationalization that survived #357 |
| #358 | opt-in experimental raster+zarr image — separate tag and endpoint, inert here (`EXTRA_DUCKDB_EXTENSIONS` unset on the default image) |
| #351 | guidance-validation runbook fix |

## Gate result (dev @ 19b3c709, regression tier, 3 trials × 2 models)

| model | t1 | t2 | t3 |
|---|---|---|---|
| z-ai/glm-5.2 | 33.3 ✅ | 33.3 ✅ | 33.3 ✅ |
| deepseek/deepseek-v4-flash-0731 | 33.3 ✅ | 33.3 ✅ | 33.3 ✅ |

**Zero occurrences** of the pre-fix 28.8% form, and zero of the 32.1% marginal form seen in the first gate. Every trial carried the area term and applied the California mask; deepseek stated the ratio explicitly as `100 × SUM(frac × (w1+w2) × cell_area) / SUM(frac × cell_area)`.

Worth recording: 33.3 is only reachable *with both* corrections — unweighted+mask gives 32.07, weighted-without-mask 29.89 — so the value corroborates the method independently. But 32.07 rounds inside the cell's ±3pt tolerance, so tolerance alone never separates a correct method from a lucky one; this cell needs a method audit.

Baseline unchanged: Mojave 7,373,400 · GAP 1+2 26,471,500 · hardwood woodland 13.6% · California 26.08–26.1% · BioRank 21.18–21.2 · channelized 22.5% (last two matching the 2026-08-03 run).

## Deploy steps after merge

```
kubectl apply -f k8s/deployment.yaml
kubectl rollout restart deployment/duckdb-mcp -n biodiversity
kubectl -n biodiversity get pods -l app=duckdb-mcp \
  -o custom-columns='NAME:.metadata.name,IMAGE:.status.containerStatuses[0].imageID'   # all 6 must converge
```

## Note

`geo-agent-benchmark` gold for `ca-cwhr13-pct-conserved` was corrected in geo-agent-benchmark#13 — it had encoded the same unweighted formula as the guidance, so the suite could not have caught this. ~~data-workflows#523 (res-9/res-8 STAC rollup recipe) is still open and contradicts the shipped guidance at the per-dataset layer; it does not block this promotion but should land soon after.~~ **Stale as written:** the per-dataset STAC text was corrected by data-workflows#524 on 2026-08-07, before this PR was opened, and #523 is closed as superseded — there is no remaining contradiction at that layer. What #524 also did was drop `nland` from the weights assets, which is what puts this promotion on hold (#364).

### DIFF
--- a/k8s/deployment.yaml
+++ b/k8s/deployment.yaml
@@ -23,7 +23,7 @@ spec:
                     values: ["us-west"]
       containers:
       - name: server
-        image: ghcr.io/boettiger-lab/mcp-data-server:v0.8.12@sha256:332d6df730c1fc66e7b2d317b121828fcb3785339a3a43b49789c3e22e6e0ffd
+        image: ghcr.io/boettiger-lab/mcp-data-server:v0.8.13@sha256:1a3409d56f62bc8c9ba252a8960eb0dad16a240e24c400620eb7afae5233db7f
         imagePullPolicy: IfNotPresent
         env:
         - name: STAC_CATALOG_URL


cboettig:
Hold — this would ship a `h3-guide.md` example that errors. Filed as #364.

`h3-guide.md:326` (the coarse-resolution overlay block from #357) reads `nland` from the
`…-hex-weights-res8` asset. data-workflows#524 removed that column an hour after #357 merged, so the
example is now a binder error, not a wrong number:

```
SQL Error: Binder Error: Table "p" does not have a column named "nland"
```

Same problem in the res-9 formula at line 336 and the prose rule at line 318. The res-10 path this
PR's gate exercised (CWHR13 percent-conserved) is unaffected and correct — which is why the gate is
green: it never touches the broken branch.

The fix is a source change, not a method change: `land_area_km2` on `ca30x30-ecoregion`'s
`ecoregion-hex-res9`/`-res8` rollups, which gives 26.135% identically and matches what the
`-hex-weights-res9`/`-res8` asset descriptions already document. Details and a suggested diff in #364.

One stale line in the description above: data-workflows#523 no longer contradicts the shipped
guidance. The per-dataset STAC text was corrected by data-workflows#524 on 2026-08-07, before this PR
was opened, and #523 is now closed as superseded.

cboettig:
Closing — superseded before it could be merged.

This promotes **v0.8.13**, which contains the coarse-overlay binder error #364: the example reads `nland` from `…-hex-weights-res8`, and data-workflows#522 removed that column (replacing it with `land_area_km2` on the land grid). Verified against live data:

```
Binder Error: Table "p" does not have a column named "nland"
```

Merging this would have shipped a res-8 example to production that cannot execute. The gate did not catch it because the regression tier has no res-8 question at all — see geo-agent-benchmark#21.

Replacement path: merge #365 (fixes #364), tag **v0.8.14**, promote that. Superseded by the promote PR for v0.8.14.

Note prod is still on v0.8.12 and therefore still serving the pre-#357 unweighted overlay guidance — the silent-wrong-answer bug from #356. That is the reason to move quickly on v0.8.14 rather than batching more work into it.

PR #365, opened 2026-08-08
fix(#364): coarse-overlay example weights by land_area_km2, not the removed nland

### LINKED ISSUE #364: h3-guide: the coarse-resolution overlay example reads nland from the weights asset, which no longer has it — binder error
`h3-guide.md`'s coarse-resolution overlay example reads `nland` from the `…-hex-weights-res8` /
`-res9` assets. That column was removed from those assets by data-workflows#522 (PR #524, merged
2026-08-07 23:33 UTC — one hour after #357), so the example is now a hard SQL error, not merely a
wrong number:

```
SQL Error: Binder Error: Table "p" does not have a column named "nland"
Candidate bindings: : "h0"
```

Reproduce (prod MCP, the example at `h3-guide.md:326` with the ca30x30 paths filled in):

```sql
SELECT 100 * SUM((p.w1 + p.w2) * p.nland * h3_cell_area(p.h8, 'km^2'))
           / SUM(p.nland * h3_cell_area(p.h8, 'km^2')) AS pct_conserved
FROM read_parquet('s3://public-ca30x30/conserved-areas-terrestrial-2025/hex-weights-res8/h0=*/data_0.parquet') p;
```

The res-8 weights schema is now `h8, w1, w2, w3, w4, h0`; res-9 is `h9, h8, w1..w4, h0`.

## Scope

Three places in the same section, all introduced by #357:

- **line 318** — the prose rule "`nland * h3_cell_area(h8, 'km^2')`, where `nland` counts the fine land cells inside it"
- **lines 326-327** — the SQL block, `p.nland` bound to the weights asset
- **line 336** — the res-9 variant, inline as `SUM(f.frac * p.w9 * p.nland * h3_cell_area(f.h9,…)) / …`

Line 331 ("The res-9 and res-8 weights assets carry a row for every land cell in the region, so this
join also bounds the result to land — no separate land grid is needed") also stops being true in
effect: the land grid is now where the weight lives, so it is joined regardless.

The res-10 same-resolution example (lines 303-315) is unaffected and correct as written.

## The method was right; only the column source is stale

`nland` and an exact per-cell `land_area_km2` now live on the region's land grid, which publishes
rollups at res 9 and res 8 (`ca30x30-ecoregion`, assets `ecoregion-hex-res9` / `ecoregion-hex-res8`,
columns `h9/h8, nland, land_area_km2, h0`). Both weights agree exactly, so this is a source fix, not
a method change — statewide GAP 1+2 at res 8:

| weight | result |
|---|---:|
| `nland * h3_cell_area(h8, 'km^2')` (the guide's product, `nland` from the land grid) | 26.135% |
| `land_area_km2` (published per-dataset recipe) | 26.135% |
| res-10 ground truth | 26.135% |

`land_area_km2` is the better of the two: it is the summed area of the actual res-10 land children
rather than a count times the parent's area, and it removes the `h3_cell_area()` call from the
query. It is also verbatim what the `-hex-weights-res9` / `-res8` asset descriptions document, so
using it keeps the injected guide and the per-dataset text on the same form — which was the point of
#357.

## Suggested change — one SQL block, the rest prose

Per AGENTS.md:137-139, keep this section example-light: the shape is already established by the
res-10 block above it, so the coarse case needs the weight named, not a second worked query.

Line 318, replace the trailing sentence:

> Weight each coarse cell by its **land** area — `land_area_km2`, published on the region's land-grid
> rollup at the same resolution (California: `ca30x30-ecoregion`, assets `ecoregion-hex-res9` and
> `ecoregion-hex-res8`) — since coastal and border cells are only partly land. Join the land grid and
> `LEFT JOIN` the weights asset to it.

Lines 322-330, the block becomes:

```sql
SELECT 100 * SUM(COALESCE(p.w1 + p.w2, 0) * e.land_area_km2) / SUM(e.land_area_km2) AS pct_conserved
FROM read_parquet('<ecoregion hex-res8>') e
LEFT JOIN read_parquet('<conserved-areas hex-weights-res8>') p USING (h8, h0)
WHERE <feature filter>;
```

Line 331 becomes: "The land grid has a row for every land cell in the region, so it is both the
weight and the denominator — no `h3_cell_area()` call and no separate mask."

Line 336, drop the inline res-9 formula to prose: "At res 9 read `ecoregion-hex-res9` and the
`-hex-weights-res9` weights the same way. When the coarse feature is itself a `hex-fractions` layer,
keep its `frac` in the product on both sides."

Net effect on the section: one fewer inline formula, same number of SQL blocks, ~30 words shorter.

## Interaction with #362

#362 (promote prod v0.8.12 → v0.8.13) would ship this section as-is, so prod would carry an example
that errors on any res-8/res-9 rollup. Worth holding for this fix and re-gating together — the #362
gate exercised the res-10 path (CWHR13 percent-conserved), so a green gate there does not cover the
broken branch. A gate cell that asks a res-8 or res-9 feature question would.

Also stale in #362's description: its closing note says data-workflows#523 "is still open and
contradicts the shipped guidance at the per-dataset layer." The per-dataset text was corrected by
data-workflows#524 on 2026-08-07, before #362 was opened, and #523 is now closed as superseded —
there is no remaining contradiction at that layer.

Found while verifying data-workflows#523.

cboettig:
Fixes #364.

`h3-guide.md`'s coarse-resolution overlay example read `nland` from the `…-hex-weights-res8` / `-res9`
assets. data-workflows#524 removed that column (2026-08-07, an hour after #357), so the example was a
hard SQL error on any res-8/res-9 question:

```
SQL Error: Binder Error: Table "p" does not have a column named "nland"
```

Source fix, not a method change: the weight now comes from the region's land-grid rollup
(`ca30x30-ecoregion`, assets `ecoregion-hex-res9` / `-res8`), which publishes an exact per-cell
`land_area_km2` — the summed area of the actual res-10 land children rather than a count times the
parent's area. That is also verbatim the form the `-hex-weights-res9` / `-res8` asset descriptions
document, so the injected guide and the per-dataset text stay on the same form (the point of #357).

## Verified against prod

Statewide GAP 1+2 at res 8, with the new block:

```sql
SELECT 100 * SUM(COALESCE(p.w1 + p.w2, 0) * e.land_area_km2) / SUM(e.land_area_km2)
FROM read_parquet('s3://public-ca30x30/ecoregion/hex-res8/h0=*/data_0.parquet') e
LEFT JOIN read_parquet('s3://public-ca30x30/conserved-areas-terrestrial-2025/hex-weights-res8/h0=*/data_0.parquet') p USING (h8, h0);
-- 26.1348
```

matching the res-10 ground truth (26.135%) and the `nland × h3_cell_area` product the guide used to
name. Also ran the filtered three-table shape as written (Mojave Desert as a res-8 feature): 44.87%
at res 8 against 44.70% at res 10 — the residual is the coarse feature's own footprint spill, which
is inherent to asking at res 8, not a weighting error.

The two res-8 assets are 1:1 over the same footprint (525,034 rows each, zero land cells without a
weights row), so the `LEFT JOIN` + `COALESCE` is defensive here and load-bearing at res 10.

## One deviation from the issue's suggested diff

#364 proposes a three-line block with a bare `WHERE <feature filter>` and no feature table. That
would not bind — the section's premise is a *coarse feature* overlaid on a finer layer, so the
feature has to be in the query. I kept the `feat` CTE from the current block and joined the land grid
to it:

```sql
WITH feat AS (
  SELECT DISTINCT h8, h0
  FROM read_parquet('<res-8 feature hex>')
  WHERE <feature filter>
)
SELECT 100 * SUM(COALESCE(p.w1 + p.w2, 0) * e.land_area_km2)
           / SUM(e.land_area_km2) AS pct_conserved
FROM feat f
JOIN read_parquet('<ecoregion hex-res8>') e USING (h8, h0)
LEFT JOIN read_parquet('<conserved-areas hex-weights-res8>') p USING (h8, h0);
```

Everything else follows the issue: line 318's rule names `land_area_km2` and the land-grid assets,
line 331 becomes "the land grid is both the weight and the denominator", and the res-9 / `frac`
variants at line 336 drop to prose. Net: one fewer inline formula, same number of SQL blocks.

## Relation to #362

#362 (prod v0.8.12 → v0.8.13) would ship the broken section, so it should take this commit and
re-gate. Its green gate exercised the res-10 path (CWHR13 percent-conserved), which is unaffected —
it never touched the broken branch. A gate cell asking a res-8 or res-9 feature question would.

### DIFF
--- a/h3-guide.md
+++ b/h3-guide.md
@@ -315,7 +315,7 @@ ORDER BY pct_conserved;
 
 Asking about **one** class ("what percent of hardwood woodland is conserved") is the same query with `WHERE f.whr13num = <code>` — keep the `frac × weight × area` product. Joining to the *distinct conserved cells* instead (a `SEMI JOIN` on `(h10, h0)`) counts every partly-conserved cell as fully conserved and overstates the percentage.
 
-**Feature coarser than the overlay layer — read the weights asset at the feature's own resolution.** *(Skip unless the feature's native resolution is coarser than the layer you are overlaying — e.g. a res-8 or res-9 feature against a res-10 coverage layer.)* Weights assets are published per resolution (`…-hex-weights-res9`, `…-hex-weights-res8`) and the coarser ones are already averaged over the fine cells inside each parent, so no rollup is needed. Weight each coarse cell by its **land** area — `nland * h3_cell_area(h8, 'km^2')`, where `nland` counts the fine land cells inside it — since coastal and border cells are only partly land.
+**Feature coarser than the overlay layer — read the weights asset at the feature's own resolution.** *(Skip unless the feature's native resolution is coarser than the layer you are overlaying — e.g. a res-8 or res-9 feature against a res-10 coverage layer.)* Weights assets are published per resolution (`…-hex-weights-res9`, `…-hex-weights-res8`) and the coarser ones are already averaged over the fine cells inside each parent, so no rollup is needed. Weight each coarse cell by its **land** area — `land_area_km2`, published on the region's land-grid rollup at the same resolution (California: `ca30x30-ecoregion`, assets `ecoregion-hex-res9` and `ecoregion-hex-res8`) — since coastal and border cells are only partly land. Join the land grid and `LEFT JOIN` the weights asset to it.
 
 ```sql
 WITH feat AS (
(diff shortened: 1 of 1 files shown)


cboettig:
Verified against live data before merge — this is the check that would have caught #364, and it is not something the regression tier can currently do (geo-agent-benchmark#21).

Running this PR's recipe statewide, against the res-10 ground truth for the same scope:

```sql
-- this PR's recipe at res 8
WITH feat AS (SELECT DISTINCT h8, h0 FROM read_parquet('s3://public-ca30x30/ecoregion/hex-res8/h0=*/data_0.parquet'))
SELECT 100 * SUM(COALESCE(p.w1 + p.w2, 0) * e.land_area_km2) / SUM(e.land_area_km2)
FROM feat f
JOIN read_parquet('s3://public-ca30x30/ecoregion/hex-res8/h0=*/data_0.parquet') e USING (h8, h0)
LEFT JOIN read_parquet('s3://public-ca30x30/conserved-areas-terrestrial-2025/hex-weights-res8/h0=*/data_0.parquet') p USING (h8, h0);
-- 26.135
```

| | |
|---|---:|
| this PR, res 8 | **26.135** |
| res-10 ground truth | **26.135** |
| pre-#522 `nland` count-weighting | 25.684 |
| the code on `main` today | *binder error* |

Exact to three decimals. It also drops the `h3_cell_area()` call the old form required, so the corrected block is shorter than the one it replaces.

Ready to merge from my side. Next: tag **v0.8.14** and promote by digest — #362 (promote v0.8.13) is closed as superseded, since v0.8.13 carries the #364 break.

Coordination note: further edits to this block are sequenced under #367, behind the gate-coverage work, so this should be the last patch here before that design pass.

Issue #371, opened 2026-08-11
META: open-issue evaluation — execution plan & chunk board


cboettig:
Durable execution plan for the open-issue evaluation (2026-08-11), so work proceeds in **discrete, context-clearable chunks** and independent chunks can run in **parallel**. Each chunk is self-contained: scope, branch, dependencies, validation, status. Update the status boxes here as chunks land.

**Status (2026-08-12): SHIPPED. A–H merged, G validated F on dev, and prod promoted to `v0.8.15@sha256:62febce2` (`/version` confirms, 6/6 serving pods converged; 3 pre-existing orphans out of rotation on disrupted nodes per the #374 pattern). Presence branch validated via the new `car-31` cell — glm-5.2, deepseek-v4-flash-0731, and the float `~deepseek/deepseek-v4-flash-latest` all 2/2. Remaining: the deferred #366 slice and the two upstream DuckDB fixes (#24736/#24737). #353 resolved (F steers to fast paths; car-28/29 + wetlands confirmed, zero timeouts).**

## Ground rules
- **GitHub Flow**: branch + PR per chunk, squash-merge via UI. Never commit to `main`.
- **Guidance-artifact changes** (`h3-guide.md`, `query-optimization.md`, `query-setup.md`, `assistant-role.md`, the `register_hex_tiles`/`query` docstrings) **must pass the headless matrix on dev before prod** (AGENTS.md → *Validating guidance changes*). These are marked **[gate]**.
- **Never run `server.py` locally** (OOMs the shared pod). Test on dev.
- Dev tracks `:main` by digest; validate there via `MCP_URL=https://dev-duckdb-mcp.nrp-nautilus.io/mcp`.

### A. Housekeeping — ✅ DONE
- [x] #354 closed (superseded by #363; S3 403 fixed in data-workflows#411)
- [x] #355 kept open, folded into chunk F (design pass) rather than wontfix

### B. #361 — DATE/TIMESTAMP rendering — ✅ MERGED (#370)
Branch `fix/361-date-rendering` → **#370**. Pure code fix, no gate. Full `test_server.py` green (108).
- [x] implemented + tested + PR + **merged**

### C. #256 — agg=COUNT vs pre-aggregated value column — ✅ MERGED (#372)
Positive-framing rule in the `register_hex_tiles` docstring: if the SELECT already produced a per-cell value via `GROUP BY hN`, pass `SUM`/`AVG`/`MAX`, not `COUNT`.
[...]

Issue #367, opened 2026-08-09
META: h3-guide fractional-overlay section — coordinate before editing; one design pass, not a fifth patch


cboettig:
**Coordination issue. If you are an agent or contributor about to edit the fractional-overlay / coarse-feature block of `h3-guide.md` (currently ~lines 295–340), read this first and comment here before starting.** Several open issues independently ask for changes to those same paragraphs. Working them separately is what produced the current state.

## Why this exists

That block has been rewritten four times in five weeks, and each edit either introduced or exposed the next defect:

| | change | what it caused |
|---|---|---|
| #350 | coarsening a coverage fraction is a mean over children, not `MAX` | correct, but 3–7× slower — turned car-29 into a timeout (#353) |
| #357 | weight the fractional overlay by `h3_cell_area` (#356) | correct, but models talked themselves out of it |
| #360 | rebut "the area terms cancel" | fixed that; shipped against a schema that changed underneath it |
| #365 | use `land_area_km2`, not the removed `nland` (#364) | fixes the binder error #360 shipped |

Every one of those defects was found *after* merge or *after* ship. That is the real problem, and it is what #245 exists to fix.

## Open issues that all touch this same block

- #363 — line networks get area-weighted; no length-weighted recipe (25.47 vs 27.46 on the same input)
- #354 — conserved-stream stats need a line-length metric; hex area is the wrong measure
- #355 — should coarse-feature-over-presence-layer be an explicit rule?
- #289 — steer to vector acre columns for GAP totals + proportional GAP-fraction weighting for overlays
- #353 — the proportional rollup is 3–7× slower; correctness and cost need deciding together

These are not five independent bugs. They are one question — **what is the correct weight for this overlay?** — whose answer depends on the geometry:

- **polygon / raster** → area (`h3_cell_area`, or `land_area_km2` from the land-grid rollup at res 8/9)
- **line** → length, deduped per feature (#363, #354)
- **presence-only layer** → neither; the open question in #355

Answering that once, coherently, is likely to produce *shorter* guidance than the current text, which now states the rule in three places with three different weight expressions. A fifth incremental patch will not.

## Sequencing — please respect this order

1. **Land #365 and ship it** (prod is currently on v0.8.12, still serving the pre-#357 silent bias).
[...]

Issue #245, opened 2026-06-24
Make guidance-change validation a pre-merge gate (sidecar MCP + NRP-side trigger)


cboettig:
## Problem

The new validation gate (AGENTS.md → "Validating guidance changes") currently requires **merge-then-validate**: a guidance change must reach `main` so dev (which tracks `:main`) can serve it, *then* the headless model suite runs against dev. This inverts the normal flow — **merge stops being the approval gate** and becomes a prerequisite *to* validation. We want validation to gate the merge instead (green ⇒ validated ⇒ mergeable).

First PR through the manual flow: #244 (validated by hand on dev). This issue is to design the automated pre-merge gate.

## Why naive "CI on the PR" doesn't work

- A **GitHub-hosted runner cannot auth to NRP** (no cluster creds in GitHub, and we don't want them there). So a PR workflow can't directly `kubectl` a Job onto NRP.
- The **MCP server is too heavy for a CI runner** (~16Gi, DuckDB + h3 extension + S3). It can't run on the GitHub runner.

So neither the test nor the server can live on the GitHub runner.

## Idea: serve PR guidance via a sidecar in the matrix Job (on NRP)

The matrix already runs **as a k8s Job on NRP** (`open-llm-proxy/headless/run-matrix-k8s.sh`), and in that loop the **MCP client is `run.js` inside the Job pod** — the model pods call back to `run.js`, which calls the MCP server. So the PR's guidance only needs to be reachable *from the Job pod*, not from the whole cluster.

That means we can run the PR's MCP server as a **second container in the matrix Job pod, on NRP**:

- `python server.py` is self-contained (port 8000, `/healthz`, `/version`, `/mcp`).
- Get the PR code in via plain **`git` checkout of the PR branch** (no image build needed — sidesteps that `docker.yml` only builds images on `main`/tags), or a `:pr-<n>` image if we prefer.
- Point the matrix at it: `run.js` already supports `--mcp-url` → `http://localhost:8000/mcp` (today `run-matrix-k8s.sh` has no MCP override and relies on the app repo's `config.mcp_url`).

Result: the exact PR guidance is validated **pre-merge**, no main pollution, no dev rollout, no separate deployment/teardown. The sidecar runs on NRP with NRP's S3/secrets — never on the GitHub runner.

## The actual hard part: triggering it from a PR

Bridging a PR event → an NRP Job, keeping all cluster creds on the NRP side. Two shapes:

[...]

PR #373, opened 2026-08-11
ci(#369): execute the guidance-file SQL examples in CI

### LINKED ISSUE #369: CI: execute the SQL examples in the guidance files — a broken example is invisible to the model gate
`h3-guide.md`, `query-optimization.md` and `query-setup.md` carry worked SQL examples that are injected into the `query` tool description. **Nothing checks that they run.** #364 shipped a coarse-overlay example that could not execute — it read `nland` from `…-hex-weights-res8` after data-workflows#522 moved that column — and it went out in a tagged release.

## Why the model gate cannot catch this

The natural assumption is that a question exercising the same path would have caught it. It would not. `car-19-ace-biorank5` **does** overlay a res-8 feature against the conserved-areas weights, and in the 2026-08-09 gate — run against dev at `19b3c709`, the build with the broken example — both models answered it correctly:

```
glm-5.2:  {'land_area_km2': 3, 'hex-weights-res8': 1}
deepseek: {'hex-weights-res8': 1, 'land_area_km2': 3}
```

They took the *asset name* from the guide and the *column names* from the per-dataset STAC description, which was already correct. A model that can route around a broken example will, so the gate goes green whether the example works or not. That is why geo-agent-benchmark#21 was closed: the coverage was already there, and adding more questions cannot fix this.

It also explains the low blast radius of #364 — degraded, not blocked. But "models mostly recover" is not a property to rely on, and the weakest models are the least likely to.

## Proposal

A CI job that extracts every ```sql block from the guidance files and executes it against the catalog.

The examples use placeholders (`<cwhr13 hex-fractions>`, `<ecoregion hex-res8>`, `<feature filter>`), so this needs a small committed fixture mapping placeholder → real path/predicate. Two rules keep the fixture honest:

- a block whose placeholders are **all** mapped must execute, or CI fails
- an **unmapped** placeholder is a failure, not a skip — otherwise the check quietly passes everything as examples drift (this is the same trap geo-agent-benchmark#9 flags for its own key mapping)

`EXPLAIN` is not enough on its own — it does bind columns, which is what #364 needed, but running the statement also catches a path that no longer exists. Given these are small scoped queries, running them is affordable; if cost becomes an issue, `EXPLAIN` catches the specific #364 shape.

## Why this is the right layer

The guidance files are **runtime artifacts**, not documentation — `server.py` loads them at startup and injects them. An example that does not execute is a defect in a shipped artifact, and it should be caught the way any broken code is: by running it in CI, before merge. The model gate measures whether guidance *steers*; this measures whether it is *valid*. Both are needed and neither substitutes.

Related: #367 (sequencing for this guidance area), #245 (make guidance validation a pre-merge gate — this is a cheap, deterministic first slice of that), geo-agent-benchmark#21 (closed, with the measurement above).

cboettig:
Closes #369. Chunk D on board #371 — unblocks the #367 design pass.

## Why
The guidance files are runtime artifacts (server.py injects them verbatim into the `query` tool description), but nothing checked that their SQL runs. #364 shipped a coarse-overlay example that read a moved column (`nland`, relocated by data-workflows#522) and it went out in a tagged release. The model gate can't catch this: a model that can route around a broken example does, so the gate goes green whether the example works or not.

## What
`tests/guidance_sql/run.py` extracts every ```sql block from the three guides and requires each to be **classified** in `tests/guidance_sql/fixture.py`:
- **EXECUTABLE** — a placeholder→value map. Every `<placeholder>` must be mapped; the substituted statement is **run**. An **unmapped placeholder is a failure, not a skip**.
- **FRAGMENT** — a non-standalone snippet (bare clause, CTE over an undefined relation, write-path COPY, or a real query whose datasets aren't mapped yet), declared with a one-line reason.

Enforcement that keeps it honest:
- a block in **neither** map → fail (you can't add a guide example without CI running it or an explicit fragment declaration);
- a fixture entry matching **no current block** → fail (editing a block changes its hash key, so a stale verdict can't ride along).

## Cost / posture
Anonymous reads against the public source.coop mirror / Ceph external endpoint — **no secrets, no cluster access** (same as `examples.yml`). Placeholders map to a **single `h0=` partition**, so a block binds every path and column and executes in **~1s total** — enough to catch a moved column, a renamed function, or a path that no longer resolves. It is not a result-value check (that's the model gate's job). This is the cheap, deterministic first slice of #245.

## First slice
**8 executable**, **23 fragments** (0 failures). The executable set includes the res-8 coarse overlay weighting by `land_area_km2` + `(w1+w2)` — **exactly the #364 shape**. Verified the mechanism catches it: reading `nland` off the weights asset raises `Binder Error: Referenced column "nland" not found`. Coverage grows by promoting "not yet mapped" fragments (several overlay ones get promoted in the #367 design pass, chunk F).

## ⚠️ Maintainer action
The workflow file `.github/workflows/guidance-sql.yml` (push/PR path-filtered to the guides + harness, weekly cron, `workflow_dispatch`) could not be pushed — my token lacks the `workflow` scope. It's ready; I'll post the exact YAML in a comment. Either add it directly, or run `gh auth refresh -h github.com -s workflow` for me and I'll push it.

Run locally: `python tests/guidance_sql/run.py` (needs `pip install duckdb==1.5.4`).

### DIFF
--- a/.github/workflows/guidance-sql.yml
+++ b/.github/workflows/guidance-sql.yml
@@ -1,0 +1,49 @@
+name: Guidance SQL
+
+# Executes the ```sql examples embedded in the guidance files (h3-guide.md,
+# query-optimization.md, query-setup.md) — which server.py injects verbatim into
+# the query tool description — against public data, so a broken example (a moved
+# column like #364's `nland`, a renamed function, a path that no longer resolves)
+# fails CI before it ships. The model gate cannot catch this: a model routes
+# around a broken example, so the gate goes green either way (#369).
+#
+# No secrets: reads go anonymously to the public source.coop mirror / Ceph
+# external endpoint, same posture as examples.yml. The runner installs the h3
+# community extension itself. The weekly cron catches the catalog rolling forward
+# under a still-valid-looking example.
+on:
+  push:
+    branches: [ main, master ]
+    paths:
+      - 'h3-guide.md'
+      - 'query-optimization.md'
+      - 'query-setup.md'
+      - 'tests/guidance_sql/**'
+      - '.github/workflows/guidance-sql.yml'
+  pull_request:
+    branches: [ main, master ]
+    paths:
+      - 'h3-guide.md'
+      - 'query-optimization.md'
+      - 'query-setup.md'
+      - 'tests/guidance_sql/**'
+      - '.github/workflows/guidance-sql.yml'
+  schedule:
+    - cron: '17 6 * * 1'   # Mondays 06:17 UTC — catch catalog drift
+  workflow_dispatch:
+
+jobs:
+  guidance-sql:
+    name: Execute guidance SQL examples
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
+        with:
+          python-version: '3.11'
+      - name: Install DuckDB
+        run: |
+          python -m pip install --upgrade pip
+          pip install 'duckdb==1.5.4'
+      - name: Run guidance SQL harness
+        run: python tests/guidance_sql/run.py
--- a/tests/guidance_sql/fixture.py
+++ b/tests/guidance_sql/fixture.py
@@ -1,0 +1,84 @@
+"""Classification of every ```sql block in the guidance files (#369).
+
+Keyed by `<file>:<sha1(block-text)[:10]>` — see run.py. Editing a block changes
+its key, forcing re-classification (a stale entry then fails CI).
+
+EXECUTABLE[key] = {placeholder: value} — every `<placeholder>` in the block must
+be mapped; the substituted statement is executed against public data.
+
+FRAGMENTS[key] = "reason" — an illustrative snippet that is not a standalone
+statement (a bare clause, a CTE referencing an undefined relation, a write-path
+COPY, or a real query whose datasets are not yet mapped).
+
+Placeholders map to a SINGLE `h0=` partition so a block binds every path and
+column and runs cheaply — enough to catch a moved column (#364), a renamed
+function, or a path that no longer resolves, which is what the model gate cannot
+see. It is not a result-value check (that is the model gate's job).
+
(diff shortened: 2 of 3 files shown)


cboettig:
Workflow file for a maintainer to add at `.github/workflows/guidance-sql.yml` (my token lacks `workflow` scope). Or run `gh auth refresh -h github.com -s workflow` and I'll push it onto this branch.

```yaml
name: Guidance SQL

# Executes the ```sql examples embedded in the guidance files (h3-guide.md,
# query-optimization.md, query-setup.md) — which server.py injects verbatim into
# the query tool description — against public data, so a broken example (a moved
# column like #364's `nland`, a renamed function, a path that no longer resolves)
# fails CI before it ships. The model gate cannot catch this: a model routes
# around a broken example, so the gate goes green either way (#369).
#
# No secrets: reads go anonymously to the public source.coop mirror / Ceph
# external endpoint, same posture as examples.yml. The runner installs the h3
# community extension itself. The weekly cron catches the catalog rolling forward
# under a still-valid-looking example.
on:
  push:
    branches: [ main, master ]
    paths:
      - 'h3-guide.md'
      - 'query-optimization.md'
      - 'query-setup.md'
      - 'tests/guidance_sql/**'
      - '.github/workflows/guidance-sql.yml'
  pull_request:
    branches: [ main, master ]
    paths:
      - 'h3-guide.md'
      - 'query-optimization.md'
      - 'query-setup.md'
      - 'tests/guidance_sql/**'
      - '.github/workflows/guidance-sql.yml'
  schedule:
    - cron: '17 6 * * 1'   # Mondays 06:17 UTC — catch catalog drift
  workflow_dispatch:

jobs:
  guidance-sql:
    name: Execute guidance SQL examples
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5.0.1
      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
        with:
          python-version: '3.11'
      - name: Install DuckDB
        run: |
          python -m pip install --upgrade pip
          pip install 'duckdb==1.5.4'
      - name: Run guidance SQL harness
        run: python tests/guidance_sql/run.py
```

cboettig:
Workflow added in 2a7b9c3 (now that the token has `workflow` scope). The **Guidance SQL** check should run on this PR since it touches `tests/guidance_sql/**`. Ready for review + merge — no gate (reads public data only).

Issue #363, opened 2026-08-08
h3-guide Problem 4: no recipe for a line network's conserved share, so models area-weight it — 25.47 vs 27.46 on the same input


cboettig:
`h3-guide.md` Problem 4 tells a model how to total line length and how to clip it to an AOI, but not how to compute **the share of a line network's length falling under a fractional per-cell overlay** — which is the question every "what percent of streams/trails/roads are conserved" ends in. With no recipe, models reach for the area-weighted overlay idiom that the rest of the guide teaches, and it silently answers a different question.

## Log evidence — a live gate, 1 cell in 4

`ca-30x30` regression tier, 2 models x 2 trials, prod `v0.8.13`, question `car-28-headwater-streams` (gold **27.0 ±1.5**, report-authoritative):

| model | trial 1 | trial 2 |
|---|---|---|
| `z-ai/glm-5.2` | **25.47** ← fail | 27.46 |
| `deepseek/deepseek-v4-flash-0731` | 28.2 | 28.2 |

Jobs `hmx-ca-30x30-notgt-{glm,ds}-20260808-2250` in `biodiversity`. The failing cell routed to the right collection and masked to California correctly — every app-layer rule fired. Its stated method was the whole defect:

> "The conserved percentage is the area-weighted average of `COALESCE(w1 + w2, 0)` across all headwater-stream cells, weighted by each cell's California land area (`land_area_km2` from the `ca30x30-ecoregion` hex-res8 asset)."

## Root cause — reproduced exactly

The per-cell rows are replicated per (flowline, cell), so an area weight applied to that row set weights each cell by how many flowlines touch it. Same masked input set, four ways of aggregating it:

| method | result |
|---|---|
| **length-weighted (correct)** — deduped `lengthkm` per feature, hex only for the per-cell GAP fraction | **27.46** |
| cell land-area weighted, rows *not* deduped ← the failing cell | **25.47** |
| cell land-area weighted, cells deduped | 28.24 |
| `h3_cell_area` weighted, cells deduped | 28.26 |
| unweighted cell average, cells deduped | 27.73 |

```sql
-- reproduces both the correct 27.46 and the failing 25.47 from identical input
WITH ca AS (SELECT DISTINCT h8, h0, land_area_km2
            FROM read_parquet('s3://public-ca30x30/ecoregion/hex-res8/h0=*/data_0.parquet')),
w AS (SELECT h8, h0, w1+w2 AS g12
[...]

Issue #355, opened 2026-08-05
h3-guide: should coarse-feature-over-presence-layer be an explicit rule? (now a single-variant failure, not a general gap)


cboettig:
# h3-guide: should coarse-feature-over-presence-layer be an explicit rule? Now a single-variant failure, not a general gap

**Rewritten 2026-08-05. The original filing claimed both shipping models collapse this. That was measured before `data-workflows#507`/`#508` landed and is no longer true — three of four model variants are now clean, and only the floating `deepseek/deepseek-v4-flash` fails. Deciding whether this still warrants a guidance rule is the open question; the evidence below is the current state.**

## The pattern

A feature whose native grid is coarser than a res-10 layer, where the finer layer carries only **presence** (no fraction column). The correct value for a parent cell is the mean of presence over its res-10 land children; promoting presence to the parent ("does any wetland cell fall in this res-8 cell?") overstates badly, and worse for sparse features than the GAP case ever did.

MRE — ACE `BioRankSW = 5` (native res-8) over NWI wetlands (res-10 presence), no GAP status anywhere:

```sql
WITH ace AS (
  SELECT DISTINCT h0, h8
  FROM read_parquet('s3://public-cdfw/ace/terrestrial-biodiversity-summary/hex/h0=*/data_0.parquet')
  WHERE BioRankSW = 5),
land AS (   -- res-10 CA land grid (data-workflows#507)
  SELECT DISTINCT h0, h8, h10
  FROM read_parquet('s3://public-ca30x30/ecoregion/hex/h0=*/data_0.parquet')),
wet AS (
  SELECT DISTINCT h0, h10
  FROM read_parquet('s3://public-wetlands/nwi-v2/hex/h0=*/data_0.parquet')
  WHERE state_code='CA' AND WETLAND_TYPE IN
    ('Freshwater Emergent Wetland','Freshwater Forested/Shrub Wetland','Estuarine and Marine Wetland'))
SELECT ROUND(100.0*SUM(CASE WHEN w.h10 IS NOT NULL THEN 1 ELSE 0 END)/COUNT(*),2) AS pct_wetland
FROM ace a JOIN land l USING (h0,h8) LEFT JOIN wet w USING (h0,h10);
-- 1.74      correct: mean of presence over res-10 land children
-- 18.86     wrong:   presence promoted to the res-8 cell  (10.8x)
```

## Current measured state — prod, 2 trials each

[...]

PR #379, opened 2026-08-12
fix(#378): disable DuckDB statistics_propagation (S3 join crash)

### LINKED ISSUE #378: DuckDB statistics_propagation crashes SEMI/INNER joins over S3 parquet with UINT64 zonemap stats (INTERNAL SetMin/SetMax)
## Summary

On the pinned DuckDB (**1.5.4**, and confirmed on **1.5.3**), a `SEMI`/`INNER` join whose probe side is a hex parquet read **over S3** raises, at plan time:

```
INTERNAL Error: SetMin or SetMax called with Value that does not match statistics' column value
```

It is the **`statistics_propagation`** optimizer computing an empty/inverted parquet zonemap intersection for a row group (multi-row-group, `UINT64`-key-sorted probe vs. a partially-overlapping build side) and asserting instead of pruning. **S3-transport-specific**: the identical bytes read from a local path or over plain `http://` do **not** crash.

## Why it matters here

This is the exact shape of masking/overlaying the **line-data (`nhdplus-hr/flowline`)** and **ACE×ecoregion** families by a region scope — i.e. the live `car-28` "% of headwater streams conserved" benchmark cell and the #363 / #355 guidance examples. On current `:main` those queries **crash in production** rather than returning a number.

- `nhd-flowline` SEMI JOIN `ecoregion/hex-res8` on `(h8,h0)` → crash.
- `ace ... JOIN ecoregion/hex` (res-10) 3-way → crash.
- `carbon`/`census` single-file joins do **not** crash (single row group; no disjoint zonemap).

Not bad data: `nhd.h8` is normal (~6.1e17, no nulls, within INT64). The keys are parquet logical `UINT64`; `pyarrow` shows `has_min_max=True`, while DuckDB's `parquet_metadata()` renders `stats_min/max` as `NULL` — the UINT64 zonemap path.

## Mitigation (this issue)

Add a per-connection `SET disabled_optimizers='statistics_propagation'` to the server preamble (`server.py`, alongside the `memory_limit` SET — an engine workaround, not model guidance), and mirror it in the `#369` guidance-SQL harness (`tests/guidance_sql/run.py`) so CI executes what the server runs.

**Performance impact: negligible.** Measured on/off over S3 on representative queries: −3.0% (mask-carbon-by-state) and +2.7% (res-10 overlay) — within noise. The join min/max pushdown this optimizer provides buys nothing here because h0 hive-partition pruning already prunes at the file level (and is unaffected).

## Upstream

Filing a minimal, credential-free repro against DuckDB (no prior report exists for this string / numeric case; the near-matches #22649/#24384/#23888 are all about *string* stats). Link to follow. This server-side SET is the mitigation until an upstream fix ships; revisit the pin then.

## Done when

- [ ] `server.py` disables `statistics_propagation` per connection (comment references this issue + the upstream link)
- [ ] `tests/guidance_sql/run.py` `make_connection` mirrors it
- [ ] The nhd-flowline SEMI-JOIN shape runs (add it as an executable guidance-SQL fixture so CI guards the regression)
- [ ] Upstream DuckDB issue filed and linked

Chunk H on the #371 board; unblocks the line/presence executable examples in chunk F (#367).

cboettig:
Closes #378 (chunk H on the #371 board).

## The bug

On the pinned DuckDB (**1.5.4**, confirmed on **1.5.3**), a `SEMI`/`INNER` join whose probe side is a hex parquet read **over S3** raises, at plan time:

```
INTERNAL Error: SetMin or SetMax called with Value that does not match statistics' column value
```

`statistics_propagation` computes an empty/inverted parquet zonemap intersection for a row group (multi-row-group, `UINT64`-key-sorted probe vs. a partially-overlapping build side) and asserts instead of pruning. **S3-transport-specific**: the identical bytes read from local disk or over plain `http://` do not crash.

This is not hypothetical — it crashes **real production queries**: masking the `nhdplus-hr/flowline` and ACE families by a region scope (the live `car-28` "% of headwater streams conserved" benchmark cell, and the #363 / #355 guidance examples) errors on current `:main` instead of returning a number. Single-file joins (`carbon`, `census`) are unaffected (one row group, no disjoint zonemap).

Minimal, credential-free repro (anonymous public S3):

```python
import duckdb
con = duckdb.connect(); con.sql("INSTALL httpfs; LOAD httpfs")
con.sql("CREATE SECRET ceph (TYPE S3, ENDPOINT 's3-west.nrp-nautilus.io', URL_STYLE 'path', USE_SSL true, SCOPE 's3://public-')")
p="s3://public-usgs-nhd/nhdplus-hr/flowline/hex/h0=577762574070710271/data_0.parquet"
b="s3://public-ca30x30/ecoregion/hex-res8/h0=577762574070710271/data_0.parquet"
q=f"SELECT count(*) FROM read_parquet('{p}') s SEMI JOIN (SELECT DISTINCT h8,h0 FROM read_parquet('{b}')) l ON s.h8=l.h8 AND s.h0=l.h0"
con.sql(q).fetchone()                                      # -> InternalException
con.sql("SET disabled_optimizers='statistics_propagation'"); con.sql(q).fetchone()  # -> (33569,)
```

## The fix

Disable this single optimizer per connection in the `server.py` preamble (alongside the `memory_limit` SET — an engine workaround, not model guidance, so it does **not** go in `query-setup.md`), and mirror it in the `#369` guidance-SQL harness (`tests/guidance_sql/run.py`) so CI executes the same plan the server runs.

**Performance: negligible.** Measured on/off over S3 on representative queries: **−3.0%** (mask-carbon-by-state) and **+2.7%** (res-10 overlay) — within noise. The join min/max pushdown this optimizer provides buys nothing here because h0 hive-partition pruning already prunes at the file level and is independent of it.

## Verification

- `tests/guidance_sql/run.py`: 8 executed, 0 failures (with the optimizer disabled).
- `pytest tests/test_server.py tests/test_dbconfig.py tests/test_s3config.py`: 146 passed.
- The exact crashing shape (`nhd-flowline SEMI JOIN ecoregion-res8`) now returns `33569` through the harness connection.

## Follow-ups

- **Upstream DuckDB issue** — minimal repro drafted; no prior report exists for this string / numeric-UINT64 case (near-matches #22649/#24384/#23888 are all *string* stats). Link to be added. Revisit the DuckDB pin when the upstream fix ships.
- **CI regression fixture** for the nhd-flowline shape rides chunk F (#367) — its line-length example encodes exactly this join, so once F lands the guidance-SQL job guards it directly.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/server.py
+++ b/server.py
@@ -170,6 +170,22 @@ def get_isolated_db(s3_key: str = None, s3_secret: str = None, s3_endpoint: str 
                 conn.sql(stmt)
             except Exception as e:
                 print(f"⚠️ Setup statement skipped: {stmt!r}: {e}", file=sys.stderr)
+        # Work around a DuckDB `statistics_propagation` bug (#378): a SEMI/INNER
+        # join whose probe side is a multi-row-group, UINT64-key-sorted parquet
+        # read over S3 raises `INTERNAL Error: SetMin or SetMax ... does not match
+        # statistics' column value` when the build side's key range is disjoint
+        # from some row groups' zonemaps (empty intersection asserts instead of
+        # pruning). Present in DuckDB 1.5.3/1.5.4; S3-transport-specific. This hits
+        # real queries — masking the nhd-flowline / ACE families by a region scope
+        # (the car-28 line-length cell) crashes instead of returning a number.
+        # Disabling this one optimizer fixes it with no measurable perf cost (h0
+        # hive-partition pruning is independent of it). An engine workaround, not
+        # model guidance, so it lives here rather than in query-setup.md. Remove
+        # when the upstream fix ships and the pin moves past it.
+        try:
+            conn.sql("SET disabled_optimizers='statistics_propagation'")
+        except Exception as e:
+            print(f"⚠️ disabled_optimizers setup skipped: {e}", file=sys.stderr)
         # Opt-in extras (experimental image only — empty list on the default tag).
         # Warn rather than raise: a missing extension should degrade this connection's
         # capability, not fail every query the replica serves.
--- a/tests/guidance_sql/run.py
+++ b/tests/guidance_sql/run.py
@@ -96,6 +96,12 @@ def make_connection():
         "URL_STYLE 'path', USE_SSL 'true', SCOPE 's3://public-')"
     )
     con.sql("SET THREADS=4")  # external endpoint is happier at low concurrency
+    # Mirror the server preamble (server.py): a DuckDB statistics_propagation bug
+    # (#378) crashes SEMI/INNER joins over some S3 hex parquet (nhd-flowline, ACE)
+    # with `INTERNAL Error: SetMin or SetMax ...`. The server disables this optimizer
+    # per connection; the harness must too, so CI executes the same plan the server
+    # runs (else a real, server-runnable example would fail here, or vice versa).
+    con.sql("SET disabled_optimizers='statistics_propagation'")
     return con
 
 


cboettig:
Upstream DuckDB issue filed and linked: **duckdb/duckdb#24736**. Reproduces on two independent S3 backends (MinIO + Ceph RGW), never local/http — so this `SET disabled_optimizers='statistics_propagation'` is the mitigation until the upstream fix ships, at which point the DuckDB pin moves and the SET comes out (both comments in this PR reference #378, which now carries the upstream link).

Issue #384, opened 2026-08-12
Guidance context budget: audit by provenance, and gate weak-model-only guidance behind an off-by-default EXTRA_INSTRUCTIONS


cboettig:
Supersedes #293, which was filed 2026-07 and is now half-fixed and aimed at the wrong target. Fresh assessment and plan below; the raw investigation is in #293's closing comment.

## The problem

`server.py:125-128` injects `query-optimization.md` + `h3-guide.md` **verbatim** into the `query` tool description. That description is part of the `tools` block on **every** LLM call, in **every** agentic loop, for **every** consumer of this MCP.

| | #293 as filed (2026-07) | today | Δ |
|---|---|---|---|
| assembled `query` description | 29,250 ch / ~8.3k tok | **42,330 ch / ~12.1k tok** | **+45%** |
| `query-optimization.md` | — | 8,461 ch (2,303 in April) | **+44% from #380 alone** |
| `h3-guide.md` | — | 32,162 ch | — |

Two distinct problems: a **one-time refund** available from guidance that was never model-validated, and the absence of any **standing discipline**, so the refund re-accretes. The second matters more.

## What has already changed

- **The `get_stac_details` half of #293 is fixed.** #303 (shared description block) and #305 (compact parent index) landed it. Measured live on the same dataset #293 cites: `get_stac_details('pad-us-4.1-fee')` is now **11,367 ch / ~3.2k tok**, down from ~22k ch, with the 56-code `State_Nm` enum rendered once instead of twice. Residual duplication ~622 ch (5%). **Out of scope here.**
- **Model-suite validation became policy on 2026-06-21** (#229, `AGENTS.md` → "Validating guidance changes"). #293 predates it. That date is the axis this issue is built on.

## Finding 1 — size is not the signal; provenance is

The largest single block is Problem 3 (raster pixels), **8,899 ch = 21% of the whole description**. It is also the *best-validated* section in the file: only **14%** of its surviving bytes predate the validation policy (the rest is #380, #309, #350). Trimming by size would have cut the strongest material first.

## Finding 2 — 54% of the guidance was never model-validated

Splitting every section's surviving bytes by the 2026-06-21 line:

| | chars | share |
|---|---|---|
| **pre-policy** | **21,851** | **54%** |
| post-policy | 18,774 | 46% |

Sections that are **100% pre-policy** — untouched since validation existed — total **~12,975 ch ≈ 3.7k tok (31% of the description)**:

[...]

Issue #359, opened 2026-08-07
query-optimization §7: filtering no-data fixes the arithmetic but silently redefines the population — missing values are not missing at random


cboettig:
`query-optimization.md` §7 tells models to filter no-data before aggregating, which is correct for the arithmetic. It does not say that the same filter **silently redefines the population being measured**, and that missing values are rarely missing at random. That omission produced a confidently-wrong published answer in the ca-30x30 app; a partner caught it, not us.

## The general rule that's missing

`WHERE col IS NOT NULL` does not select a subset of the *question*. It selects a subset of the *data* — and incomplete columns are usually incomplete for a reason that correlates with something: geography, feature type, source vintage, resolution, the region a partner happened to process first. So the surviving rows are a biased sample, and any percentage computed from them silently answers a different question than the one asked.

This is distinct from §7's concern. §7 stops a `NaN` from poisoning a `SUM`. The missing half is: **having filtered, you no longer know what population you are describing, and the filter's bias is not visible in the result.** Both halves are needed and only one is documented. §2's example at line 51 (`WHERE l.lc_class IS NOT NULL`) currently models the pattern with no caveat attached.

## Worked example (measured, reproducible)

`usgs-nhd-streams-by-order` carries Strahler `STREAMORDER` joined from USGS's `NHDFlowlineVAA`. It is populated for 11.4% of flowlines nationally, because USGS computed it for only some subbasins. A model asked "what share of California's streams by order is conserved?" wrote structurally perfect SQL whose only flaw was `WHERE STREAMORDER IS NOT NULL`:

| | |
|---|---|
| California in-network flowline | 1,083,964 km |
| …with a usable stream order | 94,505 km (8.7%) |
| …**of which HUC4 1801 (Klamath)** | 92,393 km — **97.8%** |

The model reported a statewide figure that was one basin, with GAP 3+4 at 42.9% against a statewide 25.5% — because that basin is heavily national forest. Nothing in the result hinted at it.

**The missingness was structured on two axes at once**, which is why a geography-only framing would be too narrow:

- **Region:** 15 of 22 HUC2 regions are at *literally zero* populated flowlines (millions of rows each, not rounding). Best-covered is HUC2 17 at 66.1%.
[...]

Issue #168, opened 2026-06-08
h3-guide.md: document distance-between-hexes (h3_great_circle_distance takes lat/lng, not cell indices)


cboettig:
## Symptom

In the bosl-high-seas PIHMNM analysis, **two different models independently** reached for `h3_great_circle_distance` to compute distances between hexes and passed **H3 cell indices** instead of coordinates:

- glm-5: `SQL Error: Binder Error: No function matches the given name and argument types 'h3_great_circle_distance(UBIGINT, UBIGINT)'`
- kimi: `... 'h3_great_circle_distance(UBIGINT, UBIGINT, STRING_LITERAL)'`

(Evidence: open-llm-proxy logs, origin `high-seas.nrp-nautilus.io`, 2026-06-07 — glm-5 ~mc12, kimi ~mc8.) Each burned a tool-call round recovering.

## Root cause

`h3-guide.md` documents `h3_cell_to_lat(hN)` / `h3_cell_to_lng(hN)` for getting coordinates (e.g. for map zoom), and tells models **never** to use `ST_*` spatial functions — but it gives **no pattern for computing distance between two hexes**. That's a common need (here: nautical-mile no-take / fishing-buffer zones measured from island centroids). With no guidance, models guess `h3_great_circle_distance(cellA, cellB)`, which has no matching signature.

## Proposed fix

Add a short subsection to `h3-guide.md`, e.g. **"Distance between hexes"**:

> `h3_great_circle_distance` takes **coordinates**, not cell indices: `h3_great_circle_distance(lat1, lng1, lat2, lng2, 'km')`. To measure between two H3 cells, convert each to its center first:
>
> ```sql
> h3_great_circle_distance(
>   h3_cell_to_lat(a), h3_cell_to_lng(a),
>   h3_cell_to_lat(b), h3_cell_to_lng(b),
>   'km'
> )
> ```

## Verification (confirmed via duckdb-geo MCP against the live H3 extension)

```sql
SELECT
  h3_great_circle_distance(0.0, 0.0, 0.0, 1.0, 'km') AS coords_form_km,        -- 111.195
  h3_great_circle_distance(
    h3_cell_to_lat('8029fffffffffff'), h3_cell_to_lng('8029fffffffffff'),
    h3_cell_to_lat('8003fffffffffff'), h3_cell_to_lng('8003fffffffffff'),
    'km') AS cell_to_cell_km;                                                   -- 4411.72
```

Both return sensible values; the cell-index form errors as above.

Issue #104, opened 2026-04-29
h3-guide: add worked example for h3_cell_to_lat/lng + warn against literal copy-paste


cboettig:
## What's failing

`h3-guide.md:18` mentions `h3_cell_to_lat(hN)` / `h3_cell_to_lng(hN)` in a single sentence with no example. Models reading this then fall into a predictable trap: after fetching `h8` from a prior query, they substitute the **displayed** value (rendered in scientific notation in markdown table output) as a numeric literal in the next query.

Observed today on `high-seas.nrp-nautilus.io` with `qwen3` answering "what is the largest MPA? fly to it":

1. Query returned `h8 = 6.15323e+17` (markdown table render of a BIGINT).
2. Next query: `SELECT h3_cell_to_lat(6.15323e+17) AS lat, h3_cell_to_lng(6.15323e+17) AS lng FROM ...`
3. DuckDB error:

```
Binder Error: No function matches the given name and argument types 'h3_cell_to_lat(DOUBLE)'.
        Candidate functions:
        h3_cell_to_lat(VARCHAR) -> DOUBLE
        h3_cell_to_lat(UBIGINT) -> DOUBLE
        h3_cell_to_lat(BIGINT) -> DOUBLE
```

The model then wrapped the literal in `CAST(... AS BIGINT)` to recover, but the cast was unnecessary — the real fix is referencing the column. (And even with a cast, the literal `6.15323e+17` has already lost precision — at H3 res-8 the integer ID truncated to 6 significant digits is no longer the original cell.)

A separate qwen3 session minutes later, asked the same question shape against `mpa-candidates`, used the column-reference pattern (`AVG(h3_cell_to_lat(h8))`) and succeeded first try. The difference was knowing to reference the column, not the displayed value.

## Root cause

`h3-guide.md` advertises the function but gives no canonical pattern. Models infer one — sometimes correctly (column reference), sometimes incorrectly (literal copy-paste).

## Proposed fix

Add a worked example to `h3-guide.md` for the most common use case (compute a feature centroid for `fly_to`), plus a one-line gotcha:

```markdown
### Coordinates from H3 cells

To get coordinates from hex parquet (e.g. to supply a `fly_to` center), call
`h3_cell_to_lat` / `h3_cell_to_lng` on the cell **column** in a single query.
For a feature that spans many hexes (a polygon), average:

\`\`\`sql
SELECT AVG(h3_cell_to_lat(h8)) AS lat,
[...]

PR #386, opened 2026-08-13
guidance(#384): record provenance on every guidance section, stripped at injection

cboettig:
Stage 2 of #384 — **audit, no removals.** Nothing the model sees changes.

## What lands

All 28 sections of `h3-guide.md` and `query-optimization.md` now carry a machine-readable provenance line:

```
### Problem 4 — Lines: per-segment columns and AOI boundaries
<!-- prov: issue=#153,#155,#363,#367 models=deepseek-v4-flash-0731,glm-5.2 added=2026-08-12
     cell=line-length-not-hex-area,stream-order-class-boundary tier=core -->
```

`models` is the field this was really for. It answers *"is this rule still earning its place, or was it written for a model we no longer run?"* — which currently takes a blame dig plus a dozen closed-issue reads.

Reconstructed from `git blame` + the motivating issues. **Where an issue records no model, the field says `unrecorded`** rather than inventing one. An honest gap is what justifies a later demotion; a fabricated attribution would launder it.

## Zero token cost

`strip_prov()` removes these before injection (`server.py`), so the injected text is **byte-identical** to before — verified against HEAD, 38,114 chars either way. Harness unchanged: 11 executed, 0 failures.

They live *next to the rule* rather than in `docs/` precisely so they cannot drift from it — the failure mode of every out-of-band provenance register.

## The audit, by injected weight

| | chars | share |
|---|---|---|
| covered by ≥1 benchmark cell | 29,545 | **71%** |
| `cell=none` — a rule with no cell | 7,061 | **17%** |
| `cell=background` — conceptual frame, no trap possible | 4,602 | 11% |
| `models=unrecorded` (orthogonal) | 17,597 | 42% |

**`cell=background` is a deliberate third value**, not a dodge. The data model, resolution direction, key facts and the Problems preamble are the conceptual frame a model needs to read the rules at all — "no cell" is *correct* for them and they are not demotion candidates. Without separating them the audit would read as though 28% of the guidance were unjustified, which would be wrong and would have sent Stage 3 after the wrong material.

## The `cell=none` set — what Stage 3 must answer for

| chars | models | section | issue |
|---|---|---|---|
| 1,537 | unrecorded | q-opt §9 no-data redefines the population | #359 |
| 1,292 | unrecorded | Scoping by name or feature id | #163 |
| 802 | **qwen3** | Coordinates from H3 Cells | #104 |
| 775 | unrecorded | Diagnostic: rows-per-hex | #3 |
| 683 | **qwen3** | q-opt §3 no DESCRIBE+LIKE | #108 |
| 637 | unrecorded | Generating Output Files | #60 |
| 516 | glm-5, kimi | Distance Between Hexes | #168 |
| 439 | unrecorded | q-opt §5 GeoParquet geometry | #48 |
| 380 | unrecorded | q-opt §6 Apostrophes | — |

Two observations worth carrying into Stage 3:

- **The two qwen3-era rules** (#104, #108 — 1,485 ch combined) target a model now in the `open` set only, not `standard` or `shipping`. These are the clearest `tier=extra` candidates.
- **§9 is the largest `cell=none` block and it is three weeks old**, from #380/#359 — post-policy guidance that shipped without a cell. Grow-on-fix is policy (AGENTS.md); this one slipped. That is an argument for the Stage 4 CI gate, not for demoting §9.

`tier=core` is set on everything here because that is the truth today. Stage 3 flips the demotions, and each flip is then a one-word diff that a reviewer can actually see.

Refs #384. No behaviour change, no gate needed — the injected bytes are identical.

### DIFF
--- a/h3-guide.md
+++ b/h3-guide.md
@@ -1,4 +1,5 @@
 # H3 Geospatial Indexing
+<!-- prov: issue=#49 models=unrecorded added=2026-05-28 cell=background tier=core -->
 
 **Most datasets have H3 hex versions.** Always use them for spatial operations instead of GeoParquet geometry columns.
 
@@ -8,6 +9,7 @@ When a dataset appears in the STAC catalog as GeoParquet, a hex-indexed version 
 If you browse the catalog and only find a GeoParquet with no hex equivalent, **say so** rather than falling back to spatial predicates. A missing hex version is a data pipeline gap (not something to work around silently).
 
 ## The H3 Data Model
+<!-- prov: issue=#92,#93 models=unrecorded added=2026-04-21 cell=background tier=core -->
 
 All datasets are already stored as H3 hex parquet in the STAC catalog — no conversion is needed. Understanding the origin of each dataset explains the structure you will encounter when you query it.
 
@@ -18,6 +20,7 @@ All datasets are already stored as H3 hex parquet in the STAC catalog — no con
 All spatial operations are hex joins — two datasets overlap wherever their hex IDs match. **Never use `ST_Within`, `ST_Intersects`, `ST_Centroid`, or any spatial function.** For coordinates (e.g. to supply a map zoom), use `h3_cell_to_lat(hN)` and `h3_cell_to_lng(hN)`.
 
 ## Resolution Direction
+<!-- prov: issue=#45 models=unrecorded added=2026-04-06 cell=background tier=core -->
 
 **Higher H3 resolution numbers are finer (smaller cells); lower numbers are coarser (larger cells).** h0 is the coarsest (~1000 km edge length); h15 is the finest. A higher-resolution cell is always a *child* of a lower-resolution cell — never the reverse.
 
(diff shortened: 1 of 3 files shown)


cboettig:
## Correction: §9 is instance-guarded, not uncovered

Asked what this block actually is, and the answer corrected the audit.

**What §9 is.** The selection-bias half of the no-data rule, from #359, shipped in #380. §7 stops a NaN poisoning a `SUM`; §9 is the other half — `WHERE col IS NOT NULL` **replaces the population**, and missing values are rarely missing at random. It came from a confidently-wrong published answer in the ca-30x30 app that **a partner caught, not us**: asked what share of California's streams by order is conserved, a model wrote structurally perfect SQL whose only flaw was `WHERE STREAMORDER IS NOT NULL`. Base NHD populates that column for 8.7% of California — and **97.8% of the usable length is one basin** (HUC4 1801, Klamath). It reported a statewide figure that was one heavily-national-forest basin: GAP 3+4 at 42.9% against a statewide 25.5%, with nothing in the result hinting at it.

**Why `cell=none` was wrong.** car-28 and car-29 *do* guard this — their `streamorder-use-nhdplus-hr-not-base-nhd` trap is exactly this failure, and car-28's note spells it out: *"a model that reaches for streams-by-order.STREAMORDER instead lands on Klamath-only numbers."* Gold 27.0 ±1.5 catches it numerically.

**Why `covered` would also be wrong.** Only that one dataset is guarded. §9's four steps are a *general* discipline that transfers to any partially-populated column, and no cell exercises them anywhere else.

So: new third value **`cell=instance-only:<trap>`**, which says exactly that and stops the audit flattering itself.

| | chars | share |
|---|---|---|
| covered | 29,545 | 71% |
| **instance-only** | 1,586 | 3% |
| cell=none | 5,524 | 13% |
| background | 4,602 | 11% |

**One thing this raises for Stage 3.** §9's steps 3–4 are largely a *reporting* discipline ("state the covered fraction", "the honest answer is that the breakdown is unavailable"). That is the **commentary/compliance axis**, not the trap axis — and the compliance vocabulary in `build_manifest.py` has seven categories, none of which covers an undisclosed coverage gap. Grading §9's general form probably needs a new compliance trap rather than another `trap:` cell. Worth deciding before Stage 3 treats it as a demotion candidate — it is one of the few rules here with a documented *external* cost.

Issue #387, opened 2026-08-13
query output renders H3 indexes in scientific notation — 18-digit cell ids truncated to 6 significant digits


cboettig:
## Every H3 index in `query` output is rendered in scientific notation, and the value is unrecoverable

Found while auditing guidance provenance for #384. Live on prod today:

```
SELECT h8, COUNT(*) n FROM read_parquet('s3://public-ca30x30/ecoregion/hex-res8/h0=.../data_0.parquet')
GROUP BY h8 LIMIT 3
```
```
|          h8 |   n |
|------------:|----:|
| 6.13762e+17 |   1 |
```

The true value is an 18-digit `UBIGINT`. Rendered to **6 significant digits**, so ~11 digits are gone — at res-8 the truncated number is not the same cell, and no cast recovers it. Anything the model or user does with a copied index is silently wrong.

## Cause — the #361 mechanism, integer edition

`server.py:389` renders via `df.to_markdown()`, i.e. pandas → `df.values` → tabulate. #361 fixed the *temporal* case this way and the comment at `server.py:366` already names the mechanism. The integer case was never covered, and it is **worse than #361 in two ways**:

1. **Not shape-dependent.** A mixed frame upcasts `df.values` to `float64` (uint64 + int64 → float64), but a *single* uint64 column renders scientific too — tabulate's default number formatting does it either way. So this is not "flips depending on the projection", it is **always**.
2. **Lossy, not just inconsistent.** #361 produced an ugly-but-correct timestamp. This produces a wrong number.

```
dtypes: {'h8': uint64, 'n': int64}   ->  df.values dtype: float64
single-column uint64 frame           ->  df.values dtype: uint64, still renders 6.13762e+17
```

## Fix — mirror #361: cast in DuckDB, not in the renderer

Only the SQL-side cast works. Measured on the same frame (`h8 UBIGINT, n INT, pct DOUBLE, tiny DOUBLE`):

| approach | h8 | side effects |
|---|---|---|
| current | `6.13762e+17` | — |
| `disable_numparse=True` | `6.137621790776689e+17` | **worse**; also turns `3` into `3.0`, left-aligns everything |
| `floatfmt=".10g"` | `6.137621791e+17` | still scientific |
| **`::VARCHAR` in SQL** | **`613762179077668863`** | **none — floats unaffected (`27.4638`, `1.5e-07` unchanged)** |

[...]

PR #390, opened 2026-08-16
guidance(#384): drop the h3-literal warning — #387 removed the defect it described [gate]

cboettig:
Follow-up to #389 (#387), first removal under #384.

## What goes

The "Coordinates from H3 Cells" section carried a warning built entirely on a tool defect:

> **Always pass the column** (`h8`, `h0`) — never copy a displayed h3 ID as a literal. Markdown table output may render the BIGINT in scientific notation (e.g. `6.15323e+17`), which DuckDB parses as DOUBLE (function rejects it) and which loses precision even when accepted.

#387 (merged as `1f7de11`) makes that false. Verified live on dev against prod, same query:

| prod (unfixed) | dev @ `1f7de11` |
|---|---|
| `6.13762e+17` | `613762311359823871` |
| `6.13762e+17` | `613762311519207423` |

Those are **two different cells** that prod renders as the same string. A displayed index is now exact and safe to copy, so the rule has nothing left to protect against.

## Deliberately partial

Only the warning paragraph goes. The `h3_cell_to_lat/lng` recipe **stays** — which function to call, and to average across a feature's cells, is capability information, not a workaround, and #387 does nothing to it.

#384 listed this whole 802-char section as a demotion candidate. Reading it closely, 337 chars were dead and 465 are working guidance. That is the same pattern the consumer sweep found repeatedly: `cell=none` marks *untested*, not *unneeded*, and the honest unit of removal is usually smaller than the section.

## Effect

**−337 chars (~−96 tok)** off every call, for every consumer. Modest, but the right kind: retired because the defect it described is gone, not because we argued the model no longer needs it. Third instance of the pattern that actually shrinks this prompt — fix the tool, then the text follows (after #385's dedup and §5's geometry sentence).

Harness unchanged: 11 executed, 0 failures.

## Gate

Prompt-artifact change, so per AGENTS.md this needs model-suite validation on dev before prod. Validation with `deepseek-v4-flash` (NRP, free) to follow on this PR — the check is that no baseline cell regresses, since the removed text has no cell of its own (`cell=none`).

Worth stating plainly about what that validation can and cannot show: the argument for this removal is **constructive, not statistical**. The warning described a rendering behaviour that no longer exists, which is verifiable directly (above) rather than by sampling model outputs. The matrix run is a regression check on everything else, not the evidence for the removal.

### DIFF
--- a/h3-guide.md
+++ b/h3-guide.md
@@ -100,12 +100,6 @@ FROM read_parquet('<hex parquet path>')
 WHERE <feature filter>;
 ```
 
-**Always pass the column** (`h8`, `h0`) — never copy a displayed h3 ID as a
-literal. Markdown table output may render the BIGINT in scientific notation
-(e.g. `6.15323e+17`), which DuckDB parses as DOUBLE (function rejects it) and
-which loses precision even when accepted. Accepted argument types are
-`VARCHAR`, `UBIGINT`, and `BIGINT`.
-
 ## Distance Between Hexes
 <!-- prov: issue=#168,#228 models=glm-5,kimi added=2026-06-21 cell=none tier=core -->
 


cboettig:
## Validation — no regression, `deepseek-v4-flash` (NRP) on dev @ `5be1180`

Regression tier, 2 trials, 15 questions across `ca-30x30` (10) and `biodiversity` (5). MCP target confirmed in both Job logs (`--- mcp: …dev-duckdb-mcp… (MCP_URL override) ---`), and dev verified converged first — `/version` stable at `5be1180c` over 6 samples, deployment digest matching `:main`, and the removed warning confirmed absent from the live `query` description while the kept `h3_cell_to_lat/lng` recipe is still present.

### Numerically graded — all pass

| cell | gold | trials |
|---|---|---|
| `car-19-ace-biorank5` | 21.1 | **21.2, 21.2** |
| `car-23-channelized-connectivity` | 22.7 | **22.5, 22.5** |
| `car-28-headwater-streams` | 27.0 ±1.5 | **27.46, 27.5** |
| `car-31-ace-wetland-presence` | 1.74 | **1.74**, 1 timeout |

`car-28` matches the 27.46% I reproduced directly against live data earlier, so the streams recipe is unaffected.

### L2 free-text — `needs-judge`, extracted headlines match gold

Not mechanically gradable by design, but the extracted numbers line up: `ca-gap12-acres` 26,471,500 (gold 26.47M), `ca-hardwood-woodland-pct` 13.6 / 13.6 (gold ~13.6), `ca-pct-conserved` 26.1 / 26.1 (gold ~26.1), `bio-ramsar-countries` 176 (gold UK 176), `ca-ecoregion-most-conserved` 7.38M (gold Mojave ~7.37M).

**Totals:** answer 7 pass / 14 needs-judge / 1 timeout; clarify 2 needs-judge; tool_call **6 pass**.

The single timeout is `car-31`, a known-slow presence cell that also timed out 1/2 on earlier runs — not attributable to this change.

### One caveat worth stating

The tool_call cells initially graded **6 fail**, which was a harness bug, not model behaviour — `collect_run.py` drops tool *result* rows, so `equals_result_field` failed closed on correct answers. Fixed in geo-agent-benchmark#28; re-collected from the same logs they are 6/6 pass. Flagging it because it briefly looked like this PR had caused a regression, and it had not.

### What this run does and does not establish

The case for the removal is **constructive**: the deleted text described a rendering behaviour that #387 verifiably eliminated (`6.13762e+17` → `613762311359823871` on the same query, prod vs dev). This run is the regression check on everything *else* — the removed text had no cell of its own. It does not, and cannot, prove the removal was safe by sampling; it shows nothing else moved.

**Ready for prod promotion** on the next release cut, together with #389.

Issue #394, opened 2026-08-16
h3-guide Problem 4's AOI-clip recipe is broken two ways: ST_Length returns degrees, and _cng_fid is not a cross-asset key


cboettig:
## Summary

Writing the AOI-clipped-mileage cell for the #384 backlog (geo-agent-benchmark#31 →
geo-agent-benchmark#34) required actually *running* the `ST_Intersection` recipe in
`h3-guide.md` Problem 4. It does not work. Two independent defects, both silent, both
reproducible against prod today:

1. **`ST_Length` returns degrees**, so the recipe's `/ 1609.344` is wrong by ~5 orders of
   magnitude — and every documented repair is broken on the deployed server.
2. **`_cng_fid` is not a shared key between a dataset's hex and GeoParquet assets**, which
   the recipe assumes by name and in a parenthetical. For `census-2024-county` the two
   numberings are entirely disjoint.

Neither is reachable from any existing benchmark cell: every *tested* line-length path uses
H3 and per-feature length columns (car-28/29). This is the #384 diagnosis restated — the
section was `cell=none`, `cell=none` meant untested, and untested here meant broken.

Three live consumers: `tpl`, `tpl-ca` and `wyoming` all carry `federal-trails-2026`, whose
collection exposes queryable `-parquet` and `-hex` assets alongside the display `-pmtiles`.

---

## Defect 1 — `ST_Length` is planar over lon/lat, and no unit-correct alternative works

The recipe as shipped (`h3-guide.md`, Problem 4, "Mileage *inside* an AOI"):

```sql
SUM(ST_Length(ST_Intersection(tg.geometry, rg.geometry)) / 1609.344) AS miles
```

Against prod (`duckdb-mcp.nrp-nautilus.io`, DuckDB v1.5.4):

```sql
SELECT trail_name,
       ROUND(length_miles, 2)                   AS stated_miles,
       ROUND(ST_Length(geometry), 4)            AS st_length,
       ROUND(ST_Length(geometry) / 1609.344, 8) AS guide_recipe_miles
FROM read_parquet('s3://public-trails/federal-trails-2026.parquet')
WHERE trail_name = 'Mountains to Sea Trail' LIMIT 3;
```

| trail_name | stated_miles | st_length | guide_recipe_miles |
|:---|---:|---:|---:|
| Mountains to Sea Trail | 0.48 | 0.0074 | 4.62e-06 |
[...]

PR #395, opened 2026-08-16
guidance(#394): fix the AOI-clip recipe — ST_Length is degrees, and the geometry-column line was backwards

cboettig:
Fixes the guidance half of #394. Three defects in one twelve-line block, all in the
"mileage *inside* an AOI" recipe in h3-guide Problem 4 — a section #384's audit had as
`cell=none`, which turned out to mean *nobody had ever run it*.

### 1. `ST_Length` returns degrees, and every conversion route is broken

The recipe divided by `1609.344`, which assumes metres. These GeoParquets store lon/lat, so
`ST_Length` returns **degrees** — a 74-mile trail reported as **0.00074 miles**.

The standard repairs don't exist on the deployed server:

| call | returns |
|---|---|
| `ST_Transform(…, 'EPSG:4326','EPSG:5070')` | `POINT (inf inf)` |
| `ST_Length_Spheroid` / `ST_Area_Spheroid` / `ST_Distance_Spheroid` | `nan` |
| `…::LINESTRING_2D` typed overload | `2.888e+71` |

Reproduced on stock DuckDB 1.5.4 **and** 1.4.1, so it's the extension build rather than our
deployment — but the consequence is ours: **no length or area in real units is currently
obtainable from geometry on this server.** That half of #394 needs its own issue and an
upstream report; it does not block this fix.

**The fix needs no conversion at all.** Take the clipped-to-whole ratio and scale the
feature's own `length_miles` — both lengths carry the same wrong unit, so it cancels:

```sql
SUM(tg.length_miles
    * ST_Length(ST_Intersection(tg.<geom>, rg.<geom>))
    / ST_Length(tg.<geom>)) AS miles
```

Which makes this recipe the same idea as the fractional-overlay one directly below it —
weight by the line's own length, use geometry only for the fraction — that `car-28`/`car-29`
already validate. The two adjacent recipes now agree in method as well as in subject.

### 2. The geometry-column line was backwards

> ~~The geometry column is `geometry` on the GeoParquets — not `geom`.~~

Measured across nine collections:

| column | collections |
|---|---|
| `geom` | census county / cd / place, ebsa, landvote |
| `geometry` | federal-trails-2026, seafloor-geomorphology |
| `SHAPE` | pad-us-4.1-combined, wwf-ecoregions-2017 |

There is no safe default, so the line now says to read the name off the schema. This is the
one defect of the three that failed loudly.

### 3. `_cng_fid` — no change, and that is the finding

My first pass here proposed weakening the `_cng_fid` join. **Withdrawn.** It is a per-file
row id doing exactly its job, and the advice to join on it is correct: 14 of 16 collections
checked agree, and every one that does had both assets written by a single build run.

`census-2024-county` and `-state` disagree because their hex was rebuilt **33 days** after
their flat (2026-04-22 vs 2026-03-20) from a GEOID-sorted read, so the positional id was
assigned twice with a sort in between. That is build staleness in two assets —
data-workflows#549 — not a property of the key, and not something to teach around.

### Verification

Run verbatim on prod against `census-2024-cd`, chosen because its `_cng_fid` agrees 444/444,
so the recipe shape is what's under test rather than the upstream bug:

| district | recipe as now written | naive `SUM(length_miles)` | old `/1609.344` |
|---|---:|---:|---:|
| CA-03 | **5,195.2 mi** | 5,629.9 (+8.4%) | 0.053 |
| CA-02 | **1,333.7 mi** | 1,554.0 (+16.5%) | 0.014 |

`419 passed` locally.

### Provenance

The section's `prov` gains `#394` and `line-length-clip-not-sum`, the trap tag on
`tplca-trail-miles-siskiyou` (geo-agent-benchmark#34) — so it stops being `cell=none` and
this block is guarded for the first time.

Net +14/−4 lines. It grows the prompt slightly, which is the right trade: the block it
replaces could not produce a correct answer.

### DIFF
--- a/h3-guide.md
+++ b/h3-guide.md
@@ -414,14 +414,14 @@ WITH cand AS (
   FROM read_parquet('<line_hex>') t
   JOIN read_parquet('<aoi_hex>') r ON t.h8 = r.h8 AND t.h0 = r.h0
 )
-SELECT rg.name_en AS aoi,
+SELECT rg.<aoi_name> AS aoi,
        SUM(tg.length_miles
-           * ST_Length(ST_Intersection(tg.<geom>, rg.<geom>))
-           / ST_Length(tg.<geom>)) AS miles
+           * ST_Length(ST_Intersection(tg.<line_geom>, rg.<aoi_geom>))
+           / ST_Length(tg.<line_geom>)) AS miles
 FROM cand c
 JOIN read_parquet('<line_geoparquet>') tg ON tg._cng_fid = c.trail_fid
 JOIN read_parquet('<aoi_geoparquet>') rg ON rg._cng_fid = c.aoi_fid
-GROUP BY rg.name_en ORDER BY miles DESC;
+GROUP BY rg.<aoi_name> ORDER BY miles DESC;
 ```
 
 **`ST_Length` returns degrees, not metres,** because these GeoParquets store lon/lat. Dividing
@@ -430,9 +430,10 @@ by `1609.344` is off by ~5 orders of magnitude, and the usual repairs are unavai
 ratio above is the way through — both lengths carry the same wrong unit, so it cancels, and
 `length_miles` supplies the real scale. Same idea as the fractional-overlay recipe below.
 
-**Read the geometry column name off the schema** (`DESCRIBE`, or the STAC `table:columns`) —
+**Read each geometry column name off the schema** (`DESCRIBE`, or the STAC `table:columns`) —
 it is `geom` on the census layers, `geometry` on trails and seafloor-geomorphology, and
-`SHAPE` on PAD-US and the WWF ecoregions. There is no safe default.
+`SHAPE` on PAD-US and the WWF ecoregions. There is no safe default, and the two sides of this
+join often differ.
 
 - **Share of a line network's length under a fractional per-cell overlay** ("what percent of streams are conserved", "what share of trail miles burned"): weight by the **line's own length**, never by cell area — the hex supplies only the per-cell fraction. Dedup the feature's `length_*` first (it is replicated on every cell it touches), take the mean overlay fraction across that feature's cells, then `SUM(length × fraction) / SUM(length)`:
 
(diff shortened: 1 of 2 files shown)


cboettig:
## Follow-up: the SQL harness had this block on its exempt list

CI caught my edit — and in doing so surfaced the reason the defect survived. The block was
in `tests/guidance_sql/fixture.py`'s `FRAGMENTS`, i.e. classified as *not executable*:

> `"h3-guide.md:acf2ad1267"`: line-mileage via ST_Intersection over line+AOI GeoParquet;
> **those AOI geoparquets are not mapped**

So there were **two** guards over this block and both were blind to it: no benchmark cell
(`cell=none`, #384) and a SQL harness that had been told to skip it. Binding alone would
have caught `tg.geometry` against a dataset whose column is `geom` — the harness just never
got the chance.

Promoted it to `EXECUTABLE`, mapped to `federal-trails-2026` × `census-2024-cd`. That pair
is chosen twice over:

- their geometry columns **differ** (`geometry` vs `geom`), so the mapping itself encodes the
  lesson rather than restating it;
- `census-2024-cd`'s `_cng_fid` agrees across its two assets (444/444), so the recipe shape
  is what's under test rather than data-workflows#549.

I also split the placeholder `<geom>` into `<line_geom>` / `<aoi_geom>` and turned the
hardcoded `name_en` into `<aoi_name>`. A single `<geom>` placeholder implied both sides of
the join share a column name, which is precisely the assumption that was wrong.

**Verified the gate bites rather than merely passes** — asserting the old `geometry` name for
the AOI side fails:

```
✗ EXECUTION FAILED h3-guide.md:989f348f87: Binder Error: Table "rg" does not have a column named "geometry"
```

`12 executed, 21 fragments, 0 failures`; `419 passed`.

Note what this gate can and cannot do, since it bears on the Stage 4 discussion in #384:
binding catches a moved column, a renamed function, a path that no longer resolves. It
**cannot** catch `/1609.344` — that is semantically wrong, not a bind error, and it needs the
trap cell. The two guards are complementary, which is an argument for wiring both into the
Stage 4 gate rather than picking one.

PR #397, opened 2026-08-16
k8s(dev): converge dev onto current :main (5d01287)

cboettig:
Rolls dev onto the `:main` build carrying #395, so the guidance change can be gated on dev before prod.

Digest `sha256:fcd0294…` resolved via the anonymous registry-v2 route documented in AGENTS.md (no docker CLI here); the method was sanity-checked against `v0.8.17` first and matched `k8s/deployment.yaml` exactly.

**Convergence verified** — both endpoint-backing pods report the manifest digest:

```
dev-duckdb-mcp-7bfd966f8f-q8wkj   Ready=True   …@sha256:fcd0294…
dev-duckdb-mcp-7bfd966f8f-4clxb   Ready=True   …@sha256:fcd0294…
```

`/version` reports `git_sha 5d01287`. Over the wire, the `query` description now carries the ratio form, the degrees warning and the read-the-schema line, and `/ 1609.344) AS miles` is gone (still on prod v0.8.17, as expected).

Always-on payload measured like-for-like via `tools/list`: **prod 50,811 ch → dev 51,720 ch (+909)**.

Regression tier is running on dev against tpl-ca, tpl, bosl-high-seas and ca-30x30 with NRP `deepseek-v4-flash` (free); results to follow before any prod promotion.

### DIFF
--- a/k8s/dev-deployment.yaml
+++ b/k8s/dev-deployment.yaml
@@ -32,7 +32,7 @@ spec:
         # dev's ≥2 replicas exist to catch. Bump this digest on redeploy (see
         # AGENTS.md → Rollout workflow). The tag prefix is for humans; the digest
         # is enforced. See issue #341.
-        image: ghcr.io/boettiger-lab/mcp-data-server:main@sha256:458a014dc1763d6fef21574f7889e3c8199d442640e5a2df16f343dd50c973ef
+        image: ghcr.io/boettiger-lab/mcp-data-server:main@sha256:fcd02946808587e394da975dfa523563585cba0b9e8194ad26840fe758d3219a
         imagePullPolicy: IfNotPresent
         env:
         - name: STAC_CATALOG_URL


cboettig:
## Dev gate: passed on both halves

Regression tier on dev, NRP `deepseek-v4-flash` (free, the `standard` set), 2 trials, 4 apps,
27 cells / 54 runs. Every Job's `--- mcp: ---` line confirmed dev before reading anything.

### (a) The targeted fix works — and the guidance is what steered it

`tplca-trail-miles-siskiyou`: **1,178 / 1,177.9 miles** against gold 1,177.9 (accept
1,120–1,290). Not just the right number — the model wrote the corrected recipe verbatim:

```sql
SUM(tg.length_miles * ST_Length(ST_Intersection(tg.geometry, cg.geom)) / ST_Length(tg.geometry))
```

Note `tg.geometry` **and** `cg.geom` in the same expression — the exact pairing the old
"the geometry column is `geometry` — not `geom`" line would have broken. Across both trials:
**zero occurrences of `1609.344`**, and zero attempts at `ST_Transform` / `*_Spheroid`.

The other three new cells:

| cell | answer | gold |
|---|---|---|
| `bosl-pri-eez-area` | 1,969,000 km² (both trials) | 1,968,896 km² |
| `bosl-nearest-fishing-jarvis` | 153 km / ~156 km (~84 nm) | 154 km, accept 150–160 |
| `tpl-tates-hell-acres` | 204,576 acres | 204,576 acres |

All four pass on both trials. (The `extracted` column mis-parses three of them — it grabs
"4.1" out of "PAD-US 4.1", "2024" out of the year, and a component out of a reconciliation
sum. The answers are right; the extractor is the problem, and that is pre-existing.)

### (b) No regression

Compared against `2026-08-16-regression-deepseek-v4-flash` — same model, same provider, prod
v0.8.17, i.e. the pre-change control. **All 10 shared ca-30x30 cells are grade-identical.**

`car-28-headwater-streams`, the adjacent line-length recipe and the cell most exposed to this
edit, is `pass/pass` at 27.46 / 27.5 against gold 27.0 ±1.5 — unchanged from the control.
`car-19`, `car-23`, `car-31` also unchanged. The `needs-judge` cells were `needs-judge` in the
control too.

Run recorded as `results/2026-08-16-g394-dev-regression/` with the control named in
`run.json.gate`.

### Bonus: the run reproduced data-workflows#549 live

Trial 1 followed the `_cng_fid` join and got **686 segments, 0 miles** — the stale id sent the
geometry lookup to the wrong county, so every intersection was empty. It debugged, switched to
`GEOID`, and landed on 1,177.9. Cost: 22 tool calls vs 13 on the clean trial. I've corrected
#549, where I had described that failure as silently plausible — for disjoint AOIs it is a
loud zero.

**Not promoting to prod off my own read** — flagging for the call.

PR #400, opened 2026-08-17
k8s(dev): converge dev onto current :main (e1a4a1d)

cboettig:
Rolls dev onto the `:main` build carrying #399, so the two `tier=extra` demotions can be gated on turn count before prod.

Digest sanity-checked against `v0.8.18` first (still matched `k8s/deployment.yaml` exactly) before trusting it for `:main`.

**Convergence:** both endpoint-backing pods on `sha256:16abf69…`, `/version` → `git_sha e1a4a1d`.

**The flag is off in the default image and doing its job**, verified over `tools/list`:

| | always-on | DESCRIBE+LIKE | rows-per-hex |
|---|---:|---|---|
| dev (tiered) | **50,420 ch** | absent | absent |
| prod v0.8.18 | 51,720 ch | present | present |

Regression tier running on dev across all four apps with NRP `deepseek-v4-flash`; the pre-change control is `2026-08-16-g394-dev-regression`, same model and tier with both sections present. Both demoted rules are efficiency-only, so the comparison that matters is `tool_calls` per cell — reported either way, including if turns go up.

### DIFF
--- a/k8s/dev-deployment.yaml
+++ b/k8s/dev-deployment.yaml
@@ -32,7 +32,7 @@ spec:
         # dev's ≥2 replicas exist to catch. Bump this digest on redeploy (see
         # AGENTS.md → Rollout workflow). The tag prefix is for humans; the digest
         # is enforced. See issue #341.
-        image: ghcr.io/boettiger-lab/mcp-data-server:main@sha256:fcd02946808587e394da975dfa523563585cba0b9e8194ad26840fe758d3219a
+        image: ghcr.io/boettiger-lab/mcp-data-server:main@sha256:16abf693eb28cd0652c4eac52e6751f45b7c9e29ea42af444c30ee4a2169b843
         imagePullPolicy: IfNotPresent
         env:
         - name: STAC_CATALOG_URL


cboettig:
## Gate result: both demotions are safe, but **not** by the metric the plan proposed

Regression tier on dev, NRP `deepseek-v4-flash`, 4 apps, 27 cells, 2 trials. Control is
`2026-08-16-g394-dev-regression` — same model, same tier, both sections present.

### Accuracy: unchanged

**27/27 cells grade-identical** to the control.

### Turn count: the signal is below its own noise floor

| | |
|---|---|
| total mean `tool_calls` | 235.0 → 221.0 (**−6.0%**) |
| per-cell delta | mean **−0.52**, median **0.00**, stdev **4.43** |
| direction | 10 worse, 10 better, 7 unchanged |

**I am not claiming that −6% as a win.** It is carried by a handful of large swings in both
directions, and at least one has a known unrelated cause: `tplca-trail-miles-siskiyou` reads
−11.5 because the *control* trial happened to hit the stale-`_cng_fid` detour
(data-workflows#549) and burned ~9 calls debugging it. That is run-to-run luck, not this
change. With a paired SE of 0.85, a −0.52 delta is 0.6 SE — nowhere near a result.

**This is a correction to #384's plan**, which proposed grading `extra` demotions by turn
count because `tool_call_count` is already recorded. It is recorded, but at 2 trials it is far
too noisy to resolve a rule worth a fraction of a turn. Cheap to collect ≠ sensitive enough
to decide on.

### What actually settles it: measure the behaviour the rule targets

§3 exists to stop one specific move — `DESCRIBE … WHERE column_name LIKE '%…%'` to grep for a
column. So count that, not turns:

| | before | after |
|---|---:|---:|
| `DESCRIBE … column_name LIKE` (the §3 failure) | **0** | **0** |
| **any `DESCRIBE` at all** | **0** | **0** |
| rows-per-hex profiling queries | 3 | 3 |
| `get_schema` calls | 122 | 125 |

`deepseek-v4-flash` **never emits `DESCRIBE` in either condition** — it reaches for
`get_schema`, which is exactly what §3 tells it to do. The rule was inert for this model
whether present or absent. The rows-per-hex diagnostic is likewise unchanged at 3 queries.

That is a direct, high-signal result rather than an underpowered aggregate, and it is what I
would gate future demotions on.

### The caveat that keeps this honest

This measures the **shipping** model. Both rules were added for qwen3, which is `open`-set
only — and which still receives them, because that is precisely what `EXTRA_INSTRUCTIONS=1`
is for. The demotion claim is "the shipping models are unaffected," which is what the design
asks for and what this shows. It is **not** evidence that the rules are useless; testing that
would need a qwen3 run with the flag on and off, and is a separate question from whether
prod should carry the text.

Merging this dev pin. Prod stays on v0.8.18 until you call it.

PR #408, opened 2026-08-18
ci(#402): bind the four deferred guidance-SQL blocks — the deferrals were real queries

### LINKED ISSUE #402: guidance-SQL harness: 4 FRAGMENTS are deferrals ('not yet mapped'), not non-statements — promote them to EXECUTABLE
Four entries in `tests/guidance_sql/fixture.py`'s `FRAGMENTS` are deferrals, not
non-statements. Their reasons all end in **"not yet mapped"** — meaning the SQL *is* a real
runnable query, and the only thing stopping CI from binding it is that nobody has supplied
placeholder mappings yet:

| key | reason |
|---|---|
| `h3-guide.md:39e204c12b` | GEBCO × geomorphology h6 join; two co-indexed datasets not yet mapped |
| `query-optimization.md:a941d3b637` | regions × padus × carbon 3-way; regions/padus hex paths not yet mapped |
| `query-optimization.md:48c32ed53a` | `<bucket>/<dataset>` with `_cng_fid` and `data_00.parquet`; dataset not yet mapped |
| `query-optimization.md:29b78ebc22` | generic `value`/`lc_class` columns with sentinel exclusion; dataset not yet mapped |

The module's own docstring already says what to do with them: *"To grow coverage, promote a
FRAGMENT whose reason is 'not yet mapped' to EXECUTABLE with real mappings and confirm it
runs."*

## Why this is worth a ticket rather than a TODO

**#394 was exactly this, and it cost us.** The AOI-clip recipe sat in `FRAGMENTS` with the
reason *"those AOI geoparquets are not mapped"* — a plausible, honest deferral. It shipped
with `tg.geometry` against datasets whose column is `geom`, and `ST_Length` (degrees) divided
by 1609.344. Binding alone would have caught the first defect on day one. Two guards were
nominally over that block — a benchmark cell and this harness — and both had been opted out of.

**An exemption with a good reason is indistinguishable from coverage until someone checks.**
So the deferral needs a name and a place to be tracked, which is this issue.

Promoting one is cheap: `con.sql()` binds without materialising, so a mapping only needs paths
and columns that resolve, not a query that runs fast. #395's promotion of the AOI-clip block
took one fixture entry and a pair of single-partition paths.

## Not proposed

The other 17 `FRAGMENTS` entries are genuine non-statements — bare `JOIN`/`WHERE` clauses,
CTEs over undefined relations, a write-path `COPY`, and one deliberate parse-error
illustration. Those can never bind and should stay exempt; a blanket rule against exemptions
would just add noise. The distinction that matters is **deferred** vs **not a statement**.

cboettig:
Closes #402.

Four `FRAGMENTS` entries ended in **"not yet mapped"** — real, runnable queries
waiting on placeholder mappings, not non-statements. All four now bind against
real public data.

| block | mapped to |
|---|---|
| `h3-guide` h6 parent-join | GEBCO 2025 (h8-indexed, carries `h6`) × seafloor-geomorphology (h6-indexed) |
| q-opt §2 mask-before-aggregate | Overture regions × PAD-US 4.1 combined × irrecoverable carbon |
| q-opt §2 scope-by-h0-set | IHO Maritime Boundaries — EEZ v12 |
| q-opt §7 no-data filter | CGLS-LC100 2019 hex-fractions |

**Harness: 17 executed (was 13), 17 fragments, 0 failures.** Zero deferrals remain,
so every `FRAGMENT` left is a non-statement that can never bind, and the
`check_guidance_prov.py` deferral rule now has nothing to enforce.

### Three things worth a reviewer's attention

**1. One mapping was pinned to a different partition, because binding is not enough.**
The regions × PAD-US × carbon three-way binds fine on the shared `_H0` — and returns
`NULL`. That partition holds US-CA cells and `Des_Tp='NP'` cells but no cell that is
both, so the example would have been "covered" by a join over nothing. Pinned instead
to `577199624117288959` (Yosemite / Death Valley / Joshua Tree), where it returns
6.64 Mt. The other three return real values too: the EEZ block gives 1,968,896 km²,
which is the Pacific Remote Islands EEZ — confirming that block was written against
this dataset (the five coordinates are Wake, Johnston, Kingman, Howland and Jarvis,
and the `_cng_fid` list is theirs).

**2. The EEZ block cannot be pinned to a single partition, deliberately.** Its `h0=*`
glob and `data_00.parquet` filename are literal in the block — that layout is the
thing the section teaches. Binding still only globs one prefix and reads one footer
(0.17 s), so it does not cost what a scan would.

**3. It needed one guidance edit, and that is the part to scrutinise.** No dataset in
the catalog declares a column named `value`, so q-opt §7 could never have bound as
written. (Precisely: of 266 STAC nodes, 220 declare column metadata and none declares
`value`; the 46 that declare none are container nodes — every node carrying a `/hex/`
asset declares its columns, and `read_parquet('<hex>')` reads a hex asset.)

```diff
-SELECT SUM(value) AS total
+SELECT SUM(<value>) AS total
 FROM read_parquet('<hex>')
-WHERE value IS NOT NULL AND NOT isnan(value)
+WHERE <value> IS NOT NULL AND NOT isnan(<value>)
   AND lc_class NOT IN (0, 80, 200);   -- dataset's no-data sentinels
```

`<value>` is now a placeholder like the `<hex>` beside it. `lc_class` and the
0/80/200 sentinels stay literal because they are real — they are CGLS-LC100's own
codes for unknown / permanent water / open sea, which is evidently what the block was
written against. My expectation is that this is neutral-to-better for a weak model (a
bare `value` is copyable; a placeholder is not), but that is a guess, not evidence.

**Deliberately not changed:** `h3-guide.md` has a parallel illustration using bare
`value`/`class` (the SUM/AVG-vs-MODE block at :281), so the two guides now differ in
convention. It stays a `FRAGMENT` either way — it is a two-statement block, which this
harness cannot bind — so editing it would widen a prompt-artifact change for no CI
gain. Its surrounding comment (`-- Continuous values (carbon, biomass, elevation)`)
already frames `value` as a stand-in, which §7 did not. Flagging it rather than
folding it in.

**This is a prompt-artifact change, so it needs the dev model-suite run before prod**
per AGENTS.md. §7 has a cell (`sum-amount-safe`), so the `regression` tier covers it.
Guidance budget: 40,969 ch of the 42,500 ceiling (+6 ch).

Full suite: 427 passed.

### DIFF
--- a/query-optimization.md
+++ b/query-optimization.md
@@ -147,9 +147,9 @@ dataset reserves for "no data" (e.g. land-cover classes 0, 80, 200). Exclude bot
 before aggregating:
 
 ```sql
-SELECT SUM(value) AS total
+SELECT SUM(<value>) AS total
 FROM read_parquet('<hex>')
-WHERE value IS NOT NULL AND NOT isnan(value)
+WHERE <value> IS NOT NULL AND NOT isnan(<value>)
   AND lc_class NOT IN (0, 80, 200);   -- dataset's no-data sentinels
 ```
 
--- a/tests/guidance_sql/fixture.py
+++ b/tests/guidance_sql/fixture.py
@@ -15,8 +15,10 @@ column and runs cheaply — enough to catch a moved column (#364), a renamed
 function, or a path that no longer resolves, which is what the model gate cannot
 see. It is not a result-value check (that is the model gate's job).
 
-Datasets/paths verified 2026-08-11. To grow coverage, promote a FRAGMENT whose
+Datasets/paths verified 2026-08-18. To grow coverage, promote a FRAGMENT whose
 reason is "not yet mapped" to EXECUTABLE with real mappings and confirm it runs.
+No such deferral is currently outstanding (#402 promoted the last four); every
+remaining FRAGMENT is a non-statement that can never bind.
 
 Two kinds of FRAGMENT, and only one is debt:
   * NOT A STATEMENT — a bare clause, a CTE over an undefined relation, a write-path
@@ -32,6 +34,11 @@ Two kinds of FRAGMENT, and only one is debt:
 
 # One CA partition present in every dataset used below (verified).
 _H0 = "577762574070710271"
+# A second CA partition, for the one example that needs the join to be non-empty.
+# `_H0` holds US-CA cells and PAD-US `Des_Tp='NP'` cells but no cell that is both, so
+# the q-opt §2 three-way would bind over an empty join. This one covers Yosemite,
+# Death Valley and Joshua Tree.
+_H0_PARKS = "577199624117288959"
 
 # ca30x30 conserved-areas assessment family (the #364 coarse-overlay surface)
 ECO8 = f"s3://public-ca30x30/ecoregion/hex-res8/h0={_H0}/data_0.parquet"          # h8, nland, land_area_km2, h0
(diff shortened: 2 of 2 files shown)


cboettig:
## Guidance gate: `regression` tier on dev — no regression, targeted cells clean

Run `reg-402`, dev @ `main@16c9ff6`, model `deepseek-v4-flash` (the `standard` set —
NRP, free), thinking on, 7 app slices × 2 trials = **82 cells**. Collected as
`results/2026-08-18-regression-think` in geo-agent-benchmark.

**Endpoint verified before believing anything:** all 110 MCP references across the 7
Job logs are `dev-duckdb-mcp.nrp-nautilus.io` — zero production. Both dev
endpoint-backing pods were `Ready=True` on one digest, and the served `query`
description was 40,969 ch containing `SUM(<value>)`, matching the local gate exactly.

### The targeted risk did not materialise

The one thing this change could plausibly break is a model copying the new placeholder
into SQL. **Zero occurrences of `<value>` (or any placeholder) in emitted SQL across all
7 logs**, zero binder/parser errors naming one, zero transcript errors, zero timeouts.

### `sum-amount-safe` — the cell that guards the edited section

Both cells, both trials, matched gold on every component:

| cell | gold | got |
|---|---|---|
| `tpl-boulder-funders` t1/t2 | Boulder County ~$321M dominant; NRCS ~$6.3M, USFS ~$4.9M, GOCO ~$3.1M | $321.2M / $6.3M / $4.9M / $3.1M — both trials |
| `tplca-cd16-funders` t1/t2 | ~$36.1M; Prop40 $11.75M, Prop12 $7.61M, PropTax $6.15M, LWCF ~$5.6M | $36,108,100; $11.75M / $7.608M / $6.15M / $5.638M |

One trial spontaneously noted that `amount = 0` means "recorded but unknown" — the
no-data distinction §7 exists to teach.

### Standing baseline

Zero mechanically-graded failures (13 auto-pass, 69 `needs-judge`, 0 fail). The chunk-G
quartet reproduces: car-19 **21.2** (gold 21.1), car-23 **22.5** (22.7), car-28 **27.5**
(27.0), car-31 **1.7 / 1.74** (1.74).

### One cell to read honestly, and it is not caused by this change

`car-23-channelized-connectivity` **t1** asked a clarifying question instead of
answering — *"'Channelized connectivity' appears in two datasets in the catalog, and
they'd give different answers"* — so only t2 scored. That is **pre-existing
model variance, not a regression**: the same model did the same thing on 2026-08-11 and
2026-08-16, well before this change, and the ambiguity it raises (two similarly-named
catalog datasets) has nothing to do with the no-data section that was edited.

It is arguably correct behaviour being scored as a miss, and it makes a regression cell
nondeterministic — probably worth its own issue, either to disambiguate the dataset
names or to pair the cell with a `clarify-` variant.

### Caveat on scope

69 of 82 cells are `needs-judge` (L2 free-text) and are **not** yet judged, so this is
not a full graded tally — it is the gate evidence: the targeted cells verified against
gold by hand, the auto-graded baseline clean, and no new failure mode. The `glm`
comparison pair was not run (it is now an occasional paid comparison, not the routine
gate).

Issue #406, opened 2026-08-17
AGENTS.md documents a manual dev rollout that deploy-dev.yml already automates — committed manifest drifts from the live cluster


cboettig:
## Two sources of truth for the dev image, and AGENTS.md documents the wrong one

`.github/workflows/deploy-dev.yml` converges dev automatically after every successful `main`
build — `kubectl set image` against the live Deployment, no commit. **AGENTS.md never mentions
it.** Its "Merge to `main` → redeploy dev (promote by digest)" section still gives a six-step
manual procedure: resolve the digest, edit `k8s/dev-deployment.yaml`, `kubectl apply`,
`kubectl rollout restart`, verify.

I followed that manual procedure three times this session before noticing the automation had
been doing it anyway.

### The observable symptom

```
committed k8s/dev-deployment.yaml : sha256:dd626e6…   (built from 1e91b6c)
live dev-duckdb-mcp deployment    : sha256:2723fdc…   (built from de86833)
```

Both pods are healthy and on the live digest; nothing is broken. But a reader checking what dev
runs by opening the manifest gets a stale answer, and the file will drift further with every
merge. `k8s-image-pin.yml` won't catch it — it checks that the pin *is a digest*, not that it is
*current*.

### Why it matters beyond tidiness

The manual path is not merely redundant, it is **occasionally wrong in the same direction**.
When #404 merged, no `main` build fired at all (probably a dropped webhook during the GitHub
503s), so the automation didn't run either — and `:main` still pointed at the previous build.
Following AGENTS.md there would have pinned dev to the *previous* image while believing it
carried the new guidance, and the model gate would have validated the wrong thing. I caught it
only by comparing `head_sha` across `docker.yml` runs.

So the docs need to say both things: **the automation owns dev**, and **a merge does not
guarantee an image — check the run, not the branch.**

### Options

1. **Document the automation and delete the manual steps** (keep them as the documented fallback
   for when `NRP_DEPLOY_KUBECONFIG` is unset, which the workflow already handles by skipping).
   Add the "check the run fired" warning. Smallest change, removes the trap.
2. **Have the workflow commit the digest back**, so the manifest stays the record of what dev
   runs. Heavier — a bot commit per main build — but keeps one source of truth and makes dev's
   history greppable the way prod's is.
3. **Declare `k8s/dev-deployment.yaml` a floor, not a record**, and say so in a comment in the
   file itself, so nobody reads it as current.

[...]

Issue #366, opened 2026-08-09
Dev deployment: build-once/promote-the-digest, and stop asking dev to be two environments at once


cboettig:
Follow-up to #341/#342. #342 stopped the silent cross-pod skew and was worth doing, but it bought coherence with manual toil — dev had already been hand-pinned twice in the 18 days before it merged, which is the same decay that produced #341. This issue is about the layer underneath: **dev keeps breaking because it has been asked to be two different environments, and because the image it runs is not the image prod ships.**

Three separate defects, each independently verifiable.

## 1. Dev has two contradictory mandates, both written in AGENTS.md

> **AGENTS.md:10** — "**The dev server is for development: deploy your branch there and test against it.**"
> **AGENTS.md:81** — "`:main` — moving; rebuilt on every push to `main` and by the weekly cron. **dev** tracks this."
> (also **:19** and **:264** — "dev tracks `:main`")

These cannot both hold. If dev auto-tracks `main`, the next merge — or the Monday 06:00 UTC cron — stomps whatever branch someone was testing. If dev holds a branch, it isn't "ahead of prod," it's sideways of it, and it isn't a canary for anything.

Every deploy strategy tried so far (`:main`+`Always`, then digest pinning) has been an attempt to satisfy both mandates at once. That is why none of them held.

## 2. The documented branch workflow is impossible — CI builds no branch images

```yaml
# .github/workflows/docker.yml
on:
  push:
    branches: [main]
    tags: ['v*']
  schedule: ...
  workflow_dispatch:
```

There is no `pull_request` trigger and no non-`main` branch trigger, so **no image for your branch ever exists.** "Deploy your branch there and test against it" cannot be done as written; the only route onto dev is to merge to `main` first, which defeats the point of testing before merge. Hand-pinning dev has been the workaround, and hand-pinning is what decays.

**Latent hazard in the same file** (checked — has *not* happened in the last 40 runs, so this is prevention, not diagnosis): a `workflow_dispatch` run from a feature branch falls into the `*)` case, which tags only `$repo:main`. Dispatching a build from a branch would therefore publish that branch's code to the `:main` tag.

## 3. The artifact validated on dev is not the artifact promoted to prod

This is the one that matters most, and it is directly demonstrable. Commit `19b3c70` exists as **two different images**:

```
:main   build of 19b3c70   -> sha256:f9be5dae…    ← what dev ran
[...]

PR #409, opened 2026-08-18
k8s(prod): promote v0.8.21 (guidance-SQL coverage; q-opt §7 placeholder)

cboettig:
Promotes **v0.8.21** (`16c9ff6`), shipping #402 via #408.

`k8s/deployment.yaml` only — per #406/#407 the dev manifest digest lags the cluster by
design and is not bumped here.

## What ships

The harness change is test-only. The single prompt-artifact change is one token in
`query-optimization.md` §7: `SUM(value)` → `SUM(<value>)`, because no dataset in the
catalog declares a column named `value`, so that block could never have bound as written.
Always-on payload 40,963 → **40,969 ch** (+6) against the 42,500 ceiling.

## Gate (AGENTS.md)

Dev ran `:main@sha256:cd700522` = `16c9ff6` on both endpoint-backing pods, serving a
40,969 ch `query` description containing `SUM(<value>)` — sampled four times across pods.
All 110 MCP references in the Job logs point at dev, none at production.

`regression` tier, NRP `deepseek-v4-flash`, 7 apps / 41 questions / **82 cells** × 2
trials — `results/2026-08-18-regression-think`:

- **The targeted risk did not materialise.** Zero occurrences of `<value>` or any
  placeholder in emitted SQL across all 7 logs; zero binder/parser errors naming one;
  zero transcript errors; zero timeouts.
- **`sum-amount-safe`** — the cell guarding the edited section — matched gold on every
  component, both cells × both trials: `tpl-boulder-funders` $321.2M / $6.3M / $4.9M /
  $3.1M; `tplca-cd16-funders` $36,108,100 with Prop40 $11.75M, Prop12 $7.608M, PropTax
  $6.15M, LWCF $5.638M.
- **Zero mechanically-graded failures** (13 auto-pass, 69 `needs-judge`, 0 fail). car-19
  **21.2** (gold 21.1), car-23 **22.5** (22.7), car-28 **27.5** (27.0), car-31
  **1.7/1.74** (1.74).

## One cell scored 1/2, and it is not this change

`car-23` trial 1 asked which dataset was meant rather than answering. **Pre-existing model
variance** — the same model did the same thing on 2026-08-11 and 2026-08-16, and the
ambiguity it raises (two catalog layers both documenting "channelized connectivity") has
nothing to do with the no-data section edited here. Filed as
boettiger-lab/geo-agent-benchmark#51 with evidence that both readings give 22.35%, so the
cell should stay `mode: answer` rather than become a `clarify-` cell.

## #366 caveat, stated rather than assumed

Release **rebuilds** instead of retagging, so `:v0.8.21` is `sha256:562b27b6` while the
digest dev actually gated is `sha256:cd700522` — different bytes from the same commit
`16c9ff6`. That is #366's open sliver. Byte-identity of the served `query` description is
therefore **verified after rollout**, not inferred from the shared commit.

Rollout and verification (`kubectl apply` + `rollout restart`, then digest convergence
across all 6 endpoint-backing pods, `/version` sampled repeatedly, and a prod-vs-dev
description diff) recorded in a comment below.

### DIFF
--- a/k8s/deployment.yaml
+++ b/k8s/deployment.yaml
@@ -23,7 +23,7 @@ spec:
                     values: ["us-west"]
       containers:
       - name: server
-        image: ghcr.io/boettiger-lab/mcp-data-server:v0.8.20@sha256:a522c253c47f01403c2f59f3d9842fbadbcb01b009365247ae78ef00ebec8843
+        image: ghcr.io/boettiger-lab/mcp-data-server:v0.8.21@sha256:562b27b6181c57355d2a83a853ae2f0ca434e2e2ab03e2b83ddeee6888c90de3
         imagePullPolicy: IfNotPresent
         env:
         - name: STAC_CATALOG_URL


cboettig:
## Rollout verified — prod serving v0.8.21

`kubectl apply` + `rollout restart` completed cleanly (6/6 replicas).

**Digest convergence (endpoints are authoritative):** all **6** endpoint-backing pods of
`mcp-server-duckdb` report `sha256:562b27b6…`, matching the manifest, all
`Ready=True`. Four other pods carry older digests (`a522c253…`, `5da3f35c…`, `7a52645c…`,
`c7f7c7bc…`) but are all pod-level `Ready=False` — the disrupted-node orphans described in
AGENTS.md, out of rotation and left for NRP to GC. Not force-deleted.

**`/version`:** `v0.8.21` @ `16c9ff6`, sampled **5×**, consistent every time.

**#366 byte-identity check — the point of this exercise.** Because release rebuilds rather
than retags, prod's image digest (`562b27b6…`) is not dev's (`cd700522…`). Verified the
*served artifact* instead of trusting the shared commit:

```
dev (gated):   40969 chars  sha256=a7aed648254cdad8
prod (after):  40969 chars  sha256=a7aed648254cdad8   x3 samples
→ byte-identical
```

Prod's `query` description is now **exactly** the bytes the regression tier was run
against. Diff against prod-before is only the two intended lines:

```
-SELECT SUM(value) AS total
+SELECT SUM(<value>) AS total
-WHERE value IS NOT NULL AND NOT isnan(value)
+WHERE <value> IS NOT NULL AND NOT isnan(<value>)
```

40,963 → 40,969 ch (+6). Nothing else moved.

**Functional smoke** — ran §7's own pattern against the dataset #402 mapped it to, on prod:

```sql
SELECT COUNT(*) AS n, ROUND(SUM(frac),1) AS frac_sum
FROM read_parquet('s3://public-land-cover/cgls-lc100-2019/hex-fractions/h0=577762574070710271/data_0.parquet')
WHERE frac IS NOT NULL AND NOT isnan(frac) AND lc_class NOT IN (0, 80, 200)
```

→ `n = 37,587,375`, `frac_sum = 1.95736e+07`, matching the value the CI harness computes.

### Incidental observation, unrelated to this release

That smoke output shows `DOUBLE` aggregates rendering in **scientific notation at 6
significant figures**, while `BIGINT` renders exactly — #387 fixed integers only.
Reproducible on prod:

| expression | rendered |
|---|---|
| `19573561.06::DOUBLE` | `1.95736e+07` |
| `19573561::BIGINT` | `19573561` |
| `ROUND(19573561.06::DOUBLE, 2)` | `1.95736e+07` |
| `321200000.0::DOUBLE` | `3.212e+08` |

`ROUND()` does not help. Since acreages, areas and dollar totals are mostly `DOUBLE`, a
model reporting "$3.212e+08" instead of "$321,200,000" is the visible symptom. Pre-existing,
not caused by v0.8.21 — noting it here because this rollout is where it surfaced; worth its
own issue.

Issue #410, opened 2026-08-18
query renders DOUBLE columns in scientific notation at 6 significant figures, and a NULL double as literal `nan` — #387 fixed integers only


cboettig:
## Symptom

Reproducible on prod **v0.8.21**:

```sql
SELECT 19573561.06::DOUBLE AS as_double, 19573561::BIGINT AS as_bigint,
       ROUND(19573561.06::DOUBLE, 2) AS rounded, 321200000.0::DOUBLE AS dollars
```

| as_double | as_bigint | rounded | dollars |
|---|---|---|---|
| `1.95736e+07` | `19573561` | `1.95736e+07` | `3.212e+08` |

Integers render exactly (that is #387's fix). **Doubles are cut to 6 significant figures
and flipped to scientific notation**, and `ROUND()` does not help — it is a display-layer
problem, not a value problem, so the obvious model-side workaround fails.

Since acreage, area and dollar totals are almost all `DOUBLE`, the user-visible form is a
model reporting **"$3.212e+08"** instead of "$321,200,000". `1968895.861944109` km² (the
Pacific Remote Islands EEZ) renders `1.9689e+06`.

## Second, related defect: a NULL double prints `nan`

```
|            v |
|-------------:|
|  19573561.06 |
|       nan    |     <- this row is NULL, not NaN
| 321200000    |
```

`server.py`'s comment already identifies this hazard for the types it does handle — *"It
also stops a NULL date printing as the literal `nan`, which reads as a value"* — but floats
never reach that guard.

This one is worse than it looks in context: `query-optimization.md` §7 exists to teach
models that a literal `NaN` poisons a `SUM`. So an output `nan` is now ambiguous between
"this column holds the NaN sentinel you must filter" and "this cell is NULL", which is
exactly the distinction that guidance turns on.

## Cause

`server.py:448-461`. The per-column rewrite loop special-cases `DATE`, `TIMESTAMP*` and
`WIDE_INTS = ("BIGINT","UBIGINT","HUGEINT","UHUGEINT")`, wrapping each in
`COALESCE(..., '')`. `DOUBLE`/`FLOAT`/`DECIMAL` fall through to `rendered.append(None)`, so
they are handed to `to_markdown()` unformatted and picked up by tabulate's default
`floatfmt="g"` — 6 significant digits, scientific above ~1e6.

## Proposed fix — one more branch in the same loop

`COALESCE(printf('%.10g', "col"), '')` for float columns. Tested against the
[...]

PR #412, opened 2026-08-18
k8s(prod): promote v0.8.22 (DOUBLE rendering)

cboettig:
Promotes **v0.8.22** (`c7120a3`), shipping #410 via #411. `k8s/deployment.yaml` only.

## What ships

`query` rendered `DOUBLE` columns at six significant figures — a 26-million-acre total arrived
as `2.64715e+07`, a dollar sum as `3.212e+08`. Plus a NULL double printing the literal `nan`,
which made NULL and a genuine NaN indistinguishable — the distinction q-opt §7 asks models to
act on.

No guidance text changed, so the `query` description must stay byte-identical at 40,969 ch.

## Gate

Dev @ `main@c7120a3` (`:main` `sha256:48e57d0e`, both endpoint pods converged, fix confirmed
live). `regression` tier, NRP `deepseek-v4-flash`, 7 apps / **82 cells** × 2 trials, against
this morning's `reg-402` control on the same tier, model and endpoint. All **140** MCP
references in the job logs point at dev, none at production.

**The change is visible in a model's own answer**, which is the whole argument for it:

| cell | before | after |
|---|---|---|
| `ca-gap12-acres` | "26,471,**500** acres" — the 6-sig-fig artifact | "**26,471,459** acres" — the layer's true sum |

**No baseline regression:** car-19 21.2/21.2 and car-28 27.5/27.5 unchanged; car-23 improved
to 22.5/22.5 (both trials answered rather than one clarifying — the variance tracked in
geo-agent-benchmark#51). **Zero** answers contain `e+NN` notation in either run, so the longer
numbers introduced no new failure mode.

## Two timeouts that the control did not have — and they are not this change

`car-31-ace-wetland-presence` and `wet-top10-hydrobasins-composite`. Both wetlands cells, both
the in-progress NWI rebuild: car-31's transcript shows `HTTP GET error reading
's3://public-wetlands/nwi-v2/hex/…'`, with 22 tool calls spent retrying `h0=*`, `**/*.parquet`
and `h0=0` before giving up.

Stated plainly: those two cells are **unevaluable on any build right now**, so "the gate
passed" means it passed on everything it could measure. Prod already serves the same wetlands
gap — promoting neither causes nor worsens it.

## #366 caveat

Release rebuilds rather than retags, so `:v0.8.22` is `sha256:10052de0` while dev gated
`sha256:48e57d0e` — same commit `c7120a3`, different bytes. Byte-identity of the served
description is **verified after rollout**, not inferred. Rollout record to follow in a comment.

### DIFF
--- a/k8s/deployment.yaml
+++ b/k8s/deployment.yaml
@@ -23,7 +23,7 @@ spec:
                     values: ["us-west"]
       containers:
       - name: server
-        image: ghcr.io/boettiger-lab/mcp-data-server:v0.8.21@sha256:562b27b6181c57355d2a83a853ae2f0ca434e2e2ab03e2b83ddeee6888c90de3
+        image: ghcr.io/boettiger-lab/mcp-data-server:v0.8.22@sha256:10052de08be75ca09fe2b0f10970eea1b5c1058713cf3f13d9d6ec4d473f5cd4
         imagePullPolicy: IfNotPresent
         env:
         - name: STAC_CATALOG_URL


cboettig:
## Rollout verified — prod serving v0.8.22

**Digest convergence:** all **6** endpoint-backing pods of `mcp-server-duckdb` on
`sha256:10052de0…`, matching the manifest, all `Ready=True`. Clean rollout, no orphans left
behind this time.

**`/version`:** `v0.8.22` @ `c7120a3`, sampled **5×**, consistent.

**#366 byte check** — release rebuilt (`10052de0`) rather than retagging what dev gated
(`48e57d0e`), so the served artifact was compared rather than trusted:

```
prod after:  40969 chars  sha256=a7aed648254cdad8
dev:         40969 chars  sha256=a7aed648254cdad8   → identical
prod before: 40969 chars  sha256=a7aed648254cdad8   → unchanged by this release
```

Both directions matter here. Identical to dev means prod runs the bytes the gate scored;
identical to v0.8.21 confirms **no guidance text moved**, which is what this release claimed —
any drift would have meant something unexpected shipped alongside the renderer.

**The fix, live on prod:**

```
| acres       | dollars   |                 h8 |
|:------------|:----------|-------------------:|
| 19573561.06 | 321200000 | 613762179077668863 |
```

Full magnitude on the doubles; the H3 index still exact and right-aligned, so #387 is intact
next to the per-column `disable_numparse`.

**NULL vs NaN, the second half of #410:**

```
| v   |
|:----|
| 1.5 |
|     |     <- NULL
| nan |     <- genuine NaN
```

Distinguishable for the first time. q-opt §7 tells models to filter the NaN sentinel before
aggregating; until now both rendered `nan`, so that instruction pointed at something invisible.

Closes out #410.

Issue #413, opened 2026-08-18
Release rebuilds the artifact instead of retagging it, so prod never provably runs the bytes dev gated (#366 boxes 2/3/6, now untracked)


cboettig:
Successor to #366, which closed 2026-08-16 with boxes 2, 3 and 6 unticked. Their only
remaining home is #371 — a board that declares its own job done — so this defect is
effectively untracked while still being hit on every release.

## The symptom, twice today

| release | digest dev gated (`:main`) | digest prod got (`:vX.Y.Z`) |
|---|---|---|
| v0.8.21 | `sha256:cd700522…` | `sha256:562b27b6…` |
| v0.8.22 | `sha256:48e57d0e…` | `sha256:10052de0…` |

Same commit each time. Different bytes. #366 measured it more sharply on v0.8.16: **6 of 10
layers differed**, and an unpinned `pandas` at that moment could have silently reverted the
#387 fix that release was shipping — the symptom would have looked like the bug was never
fixed. That is why `pandas` is pinned (#392): a workaround for this, not a fix.

Both promotions today were only safe because I diffed the *served* `query` description
against dev by hand. **That manual check is the compensating control for this defect**, and
it does not scale — it catches guidance drift but would not catch, say, a different DuckDB
patch release.

## Root cause: the version stamp is baked in at build time

`docker.yml:70-74` sets `app_version` from the git ref, passes it as a **build arg**
(`:89-91`), and `Dockerfile:28-30` turns it into `ENV APP_VERSION`. Because the identity is
*inside* the image, changing it requires building a new image — so `tags: ['v*']`
(`docker.yml:6`) triggers a second build of a commit that was already built and already
validated.

Everything downstream follows from that one decision: a rebuild cannot be bit-identical, so
"promote the artifact you tested" is unachievable by construction, so every promotion needs
manual equivalence checking.

## This is a known anti-pattern with a standard fix

**Build once, promote the digest.** The artifact is immutable and identity-free; release
identity is metadata attached at promotion, not baked at build.

#366's closing comment treated decoupling the version stamp as a significant prerequisite. It
mostly isn't — **`server.py:36` already reads `APP_VERSION` from `os.environ`**, and a
Kubernetes `env:` entry overrides an image `ENV` with no code change:

1. **`k8s/deployment.yaml`** — add `APP_VERSION` (and optionally `GIT_SHA`) to `env:`,
   alongside the existing digest pin. Both move together in one promotion commit, which is
[...]

PR #417, opened 2026-08-19
k8s(prod): promote v0.8.23 (register_hex_tiles display handoff)

### LINKED ISSUE #330: register_hex_tiles guidance: weak models narrate 'done' and skip add_hex_tile_layer (test current vs proposed on qwen)
## Symptom

Small open models (observed on **`qwen`** / the nimbus "dse-nimbus" endpoint, on padus → dev) frequently fail to actually put a hex layer on the map even though `register_hex_tiles` succeeded. On a cache-hit / `status:"done"` register, the model reads the returned "paste-ready MapLibre render recipe" and the docstring line *"You render those verbatim; you never choose"*, concludes the map is already updated, **narrates success in prose, and never calls the client `add_hex_tile_layer` tool** — so nothing appears.

Related, weaker forms seen the same day: the model *does* call `add_hex_tile_layer` but **hallucinates the `tile_url`** (fake hosts, all-zero hashes, `abc123def456`) instead of copying the returned `tile_url_template`, and/or passes a wrong `value_column`; the client's metadata-by-hash fetch then fails with the unhelpful `add_hex_tile_layer: could not resolve color-scale metadata (value_stats, bounds)`.

## Evidence (2026-07-15, padus, qwen)

Segmenting the day's traffic into distinct user requests: **8 hex requests → 4 plotted OK, 3 failed, 1 tool-call-leak no-op (~50%).** Representative failure — request "render global gbif richness hex map at h5 log scale": `register_hex_tiles` returned `status:done`; the model then said *"The hex tiles are built and cached… here's what was generated"* and stopped, never calling `add_hex_tile_layer`. This is post-fix traffic (dev on `main`, past #323/#327) — the arg-name (#321) and schema-title-leak (#326) failures are gone; this is a distinct, current issue.

Root cause is the **server-recipe vs client-handoff mismatch**: `register_hex_tiles` frames its output as a render recipe to apply verbatim, but on the real apps the display path is the *client* tool `add_hex_tile_layer(tile_url, value_column, value_stats)`. The docstring's framing actively steers a weak model to skip that call. (See also the server-vs-client styling split.)

## What to do

This is a **guidance change**, so it must go through the headless model-suite validation in `AGENTS.md` → *"Validating guidance changes"*. The goal is the most **concise** wording that fixes the behavior **without regressing** the baseline set — per AGENTS.md, guidance is shared context read by every call; prefer editing the existing `register_hex_tiles` docstring text over adding a parallel rule.

Concretely:

1. **Quantify the baseline rate.** Run the headless matrix (`open-llm-proxy/headless/run-matrix-k8s.sh` against `boettiger-lab/geo-agent-template`, `mcp_url`→dev) on the *current* guidance with a hex-display trap question (e.g. the h5 GBIF-richness ask above) at `TRIALS`≥5. **Include `qwen` in `MODELS`** since that is the model exhibiting the failure (it routes to the nimbus provider) — plus a spread of the standing suite (`glm-5 kimi qwen3 minimax-m2 gpt-oss`). Measure: fraction of trials where a hex layer is actually added (`add_hex_tile_layer → success:true`) vs. narrated-done-without-add and vs. hallucinated-URL.
2. **Draft proposed guidance.** Candidate direction: make the `register_hex_tiles` `status:"done"` response state plainly that the layer is **NOT yet on the map** and that the next required step is `add_hex_tile_layer(tile_url=<tile_url_template>, value_column=<...>)` — copied verbatim from the return value — rather than describing a "render verbatim" recipe. Keep it to the minimum words that move the number. Consider whether the fix belongs in the docstring, the `status:done` payload, or both (payload text is read exactly when the decision is made).
3. **Optionally** improve the `could not resolve color-scale metadata` error text to instruct "copy `tile_url_template` verbatim; do not construct the URL" — cheap, and directly addresses the hallucination sub-mode. (This is a client/error-message change, trackable separately if it lands in geo-agent.)
4. **Validate both conditions** per AGENTS.md: (a) the narrated-done-without-add failure form drops to ~zero across the models that showed it, **and** (b) no regression on the standing baseline set. A baseline regression blocks promotion even if the trap is fixed.
5. If the fix works, **add the trap question to the standing baseline** so it can't silently regress.

## Acceptance

- Headless matrix (current vs proposed guidance, same job) shows the "registered but never added the layer" rate for qwen (and any other affected model) falling from its measured baseline to ~0, with the proposed wording as short as possible.
- No baseline-set regressions.
- Trap question added to the baseline set.

## Notes / constraints (from AGENTS.md)
- Do **not** run `server.py` locally — test on dev only.
- Only the ✅ form in prompt text; no ❌ antipattern examples (small models reproduce them verbatim, #133/#134).
- Ground-truth the correct behavior yourself first so transcripts have something to check against.

cboettig:
Ships #330 via #416. **First release through the retag path (#413)** — and it delivers the
thing #366 tracked for weeks:

```
dev pods   sha256:ddc8fe69…
:main      sha256:ddc8fe69…
:v0.8.23   sha256:ddc8fe69…
```

**Prod pins the exact artifact the gate scored.** v0.8.21 and v0.8.22 each shipped a digest
dev never ran, because a release tag rebuilt the commit; both were only safe because I diffed
the served description by hand. That manual check is no longer the safety net.

`docker.yml` **did not fire on this tag** — confirmed, the last two builds are both `main`
pushes. `APP_VERSION` now comes from the Deployment, which is what made the rebuild
unnecessary; `check_k8s_image_pins.py` enforces that it matches the image tag.

## What ships

`register_hex_tiles` framed its output as a paste-ready MapLibre recipe and never named the
client tool that displays it. A weak model read `status:"done"`, concluded the map was
updated, narrated success and stopped — ~50% of hex requests on qwen. Every done response now
carries `next_step`.

## Gate

Dev @ `main@5f5cf42`. Regression tier, 7 apps / 82 cells / 2 trials (deepseek-v4-flash,
qwen3-small), then 3 trials on the biodiversity slice adding `z-ai/glm-5.2`.

| production model | hex-map cells |
|---|---|
| `deepseek-v4-flash` | 3/3 pass |
| `z-ai/glm-5.2` | 3/3 pass |

**10/10 trials across three models** called `add_hex_tile_layer` with a `tile_url` matching a
returned template — zero narrate-and-stop, zero fabricated URLs.

Delivery of the new field verified by result-length against the reg-410 control: done payloads
**1992 → 2366** median, distributions non-overlapping.

## The one gating failure was a grader defect, not a regression

`equals_result_field` compared against only the **first** `register_hex_tiles` result, so a
model that refined its SQL and re-registered was failed for handing off the correct (second)
URL. Diagnosed in `grade.py`, reported, fixed upstream (geo-agent-benchmark#66, which also
found a second instance in `_result_truncated`). Re-grading 209 cells flipped exactly the two
affected trials and nothing else.

`qwen3-small` fails that cell genuinely (`agg=MAX`, 3/3) — the canary working, and per the
gating rule not a ship blocker.

## Does NOT close #330

Every register in every run returned `"running"`, so the **cache-hit path the issue actually
reported has never been exercised**. Shipping improves the poll path; the originally-observed
failure remains unverified. Closing needs a cell whose SQL hash is pre-warmed in setup, which
is benchmark-side and shouldn't be written until the warm path is shown reproducible.

### DIFF
--- a/k8s/deployment.yaml
+++ b/k8s/deployment.yaml
@@ -23,7 +23,7 @@ spec:
                     values: ["us-west"]
       containers:
       - name: server
-        image: ghcr.io/boettiger-lab/mcp-data-server:v0.8.22@sha256:10052de08be75ca09fe2b0f10970eea1b5c1058713cf3f13d9d6ec4d473f5cd4
+        image: ghcr.io/boettiger-lab/mcp-data-server:v0.8.23@sha256:ddc8fe69a382ff6ef09624f0e861c1a7cf465007ca1c8500ea288fab6cc4fc31
         imagePullPolicy: IfNotPresent
         env:
         # Release identity is supplied HERE, not baked into the image (#413). The
@@ -34,7 +34,7 @@ spec:
         # the image's own value is the commit it was built from, which stays correct
         # through a retag.
         - name: APP_VERSION
-          value: "v0.8.22"
+          value: "v0.8.23"
         - name: STAC_CATALOG_URL
           value: "https://s3-west.nrp-nautilus.io/public-data/stac/catalog.json"
         - name: MCP_PUBLIC_BASE_URL


cboettig:
## Rollout verified — and the #413 invariant holds on its first real use

**`/version`: `v0.8.23` @ `5f5cf42`, sampled 5×, consistent.** This was the specific failure
mode to watch: the image stamps itself `main` now, so the release identity comes entirely from
the Deployment's `APP_VERSION` env. If that had not taken, prod would be reporting `main` in
production. It reports the release.

**Digest convergence — all 6 endpoint-backing pods on `sha256:ddc8fe69…`, all `Ready=True`.**
No orphans this time.

**The invariant this release exists to establish:**

```
dev pods (gated)   sha256:ddc8fe69…
prod pods (live)   sha256:ddc8fe69…   ← identical
```

Prod is running the exact bytes the regression tier scored. On v0.8.21 and v0.8.22 those were
different digests from the same commit, and equivalence had to be argued by diffing the served
tool description by hand. That check is now a spot check rather than the safety net — which
was the entire point of #366/#413.

**The fix is live**, sampled 3× on a cache-hit register:

```
status=done  cache_hit=True  next_step=present
"Tiles are built, but they are NOT on the map yet. To display them, call your client's
 map-display tool (`add_hex_tile_layer` in the geo-agent apps) with `tile_url` copied
 verbatim from `tile_url_template` above…"
```

Sampled repeatedly rather than once, because a single mid-rollout probe against a
not-yet-replaced pod produced a false negative during this change's dev verification and
nearly sent me debugging correct code.

**#330 stays open**, as stated in the PR body: the cache-hit branch has now been exercised by
me directly against prod, but never by a *model* in a graded run — every register in every
matrix run returned `"running"`. The behaviour that issue reported is still unverified.

PR #421, opened 2026-09-13
feat(#420): catalog scope is deployment config, not a tool argument

### LINKED ISSUE #420: Catalog scope should be a deployment property, not a model-supplied tool argument
## Summary

Catalog scope is currently a **per-call tool argument authored by the model** (`catalog_url`, `catalog_token`, plus an unbounded `dataset_id`). It should be a **session/deployment property** the model cannot see or set.

This is not a new direction — it is finishing the one [`docs/architecture/catalog-sourcing.md`](docs/architecture/catalog-sourcing.md) (#265) already committed to, and it repeats a refactor this stack has already done once, successfully, for credentials.

Client-side counterpart: boettiger-lab/geo-agent#354.

## The problem

A geo-agent app is configured with a specific set of collections. The agent is nonetheless able to reach **any** dataset in **any** STAC catalog, because the scope of the data universe is re-decided by the model on every call:

| Tool | Scope-carrying argument | Who decides |
|---|---|---|
| `get_stac_details` / `get_collection` | `catalog_url`, `catalog_token`, unbounded `dataset_id` | model |
| `browse_stac_catalog` | `catalog_url`, `catalog_token` | model |

Observed in production on the ca-30x30 app: the agent discovers and reports on datasets the app does not configure or display. App owners have no way to prevent it, because the capability is in the tool schema rather than in deployment config.

Client-side argument validation was considered and rejected — it re-decides per call something the deployment should state once, and it cannot bind a tool argument the model is free to supply.

## Where these parameters came from

They were added for a real need — private STAC catalogs — and the *intent* was always session scope. Tool arguments were simply the only channel the API offered.

| Where | What |
|---|---|
| #18 (`9b596b4`) | "Allow client to specify STAC catalog URL via optional `catalog_url` parameter" |
| #20 (`ea17951`) | "Support client-supplied credentials for private STAC catalogs and S3 data" |
| geo-agent#187 (Apr 2026) | `get_schema` didn't forward `catalog_url`, so the private wyoming app got `Dataset 'gye-parcels' not found` for *every* private collection |

## Why they are now vestigial

`#105`/`#107` added inline `collection` content, and the documented resolution order is **inline `collection` → `catalog_url` → server default** (`stac.py:912` for `get_dataset`, `stac.py:986-996` for `get_collection`).

Inline wins. And geo-agent now *always* injects inline for configured data:

- `app/map-tools.js:725-728` — `get_schema` sends `{dataset_id, collection}`, **no** `catalog_url` (the #187 fix was superseded by inline forwarding)
- `app/main.js:369-381` — the `injectInlineStac` argsRewriter injects inline on direct model calls too
- `app/dataset-catalog.js:66` — `catalogToken` is assigned and **read nowhere**; dead code

So for every legitimate in-app call, `catalog_url` is never read. The only live reference to it in the entire geo-agent client is `app/main.js:371`, a pass-through guard whose sole function is letting *the model's* `catalog_url` through untouched.

**The parameters now exist to serve exclusively the case we want to eliminate.** #265's own reliance map agrees: `get_stac_details` / `get_collection` are "Fallback only… in production ~always inline."

## Precedent: credentials already made this exact move

From `mcp-credential-proxy/mcp_proxy.py` in the wyoming app:

> S3 credentials are NOT handled here anymore. The replica routes `s3://private-wyoming` via a scoped DuckDB secret (its `S3_SOURCES` env), so credentials live only in the replica's env and never appear in any JSON-RPC body… This also gives `register_hex_tiles` (which has no `s3_*` params) a working private path, which the old per-request credential injection could never provide.

Credentials moved from model-visible per-request params → per-deployment env, and doing so **unlocked** a capability rather than restricting one. Catalog scope is the same parameter class and should follow it out of the JSON-RPC body.

That sidecar already intercepts and rewrites `catalog_url` in flight (`_inject()`), so a component in the right position already exists — it normalizes the host but does not constrain the value.

## Proposal

**1. Remove `catalog_url` / `catalog_token` from the model-facing schemas** of `get_stac_details`, `get_collection`, and `browse_stac_catalog`. Catalog identity becomes deployment config (`STAC_CATALOG_URL` + a token from env), exactly as S3 credentials already are. Keep inline `collection` — it is the primary contract under "carry the links."

**2. Make `browse_stac_catalog` opt-in per deployment** (e.g. an env flag, default off or default on — see open questions). By #265's own numbers it is ~1.4% of calls and "optional discovery," and it is the only tool whose *purpose* is unscoped. Apps wanting curation-only simply don't get it in `tools/list`.

**3. `get_stac_details` / `get_collection` otherwise unchanged.** Because inline already wins the resolution order and production always sends inline, configured-data lookups keep working byte-for-byte. This is the point that makes the change safe.

### What explicitly does *not* change

- The `query` path. Path-scoping for SQL is a separate, larger question — `s3config.py`'s source registry is the natural deployment-level analog when you want it, but it is out of scope here.
- Inline `collection` / embedded `children` rendering.
- Server-side default resolution for deployments that want a public catalog.

## Migration

- **Blast radius is near zero on the client**: geo-agent already sends neither parameter. Grep of `app/*.js` finds `catalog_url` only in the pass-through guard.
- A deprecation window that accepts-but-ignores the arguments (with a log line) would cover any unknown direct MCP consumers.
- **Wyoming becomes the first beneficiary.** Its replica currently sets `STAC_CATALOG_URL` to the **public** NRP catalog (`k8s/mcp-deployment.yaml`, commented "public discovery/query keep working") and reaches the private catalog via per-request `catalog_url`. So today the app holding confidential private-parcel data has the *widest* discovery surface of any deployment. Under this proposal it sets `STAC_CATALOG_URL` to its own in-cluster wyoming catalog and turns discovery off — which also makes the sidecar's `catalog_url` rewrite unnecessary.

## Verification

1. `tools/list` no longer advertises `catalog_url` / `catalog_token` on any tool.
2. `get_stac_details` with inline `collection` returns identical output to today (regression fixture).
3. With discovery disabled, `browse_stac_catalog` is absent from `tools/list`.
4. A call naming a foreign `catalog_url` resolves against the deployment's configured catalog, not the supplied one.
5. Wyoming: `get_schema` on `gye-parcels` still returns the private schema (the geo-agent#187 regression does not return).

## Open questions for whoever picks this up

- Should `browse_stac_catalog` default **on** (back-compat for the public duckdb-mcp) or **off** (safe-by-default, every deployment opts in)? Public prod arguably wants it on; every private app wants it off.
- Is a deprecation window warranted at all, or are geo-agent + the wyoming sidecar the only consumers? If so this can be a clean removal.
- Does anything rely on `catalog://list/{id}` resources for cross-catalog reads? They carry the same whole-catalog reliance as `browse`.
- Worth folding the outcome back into `docs/architecture/catalog-sourcing.md` as the settled rule ("scope is ambient, never an argument") so the next tool with a convenient `catalog_url` param doesn't reintroduce it.

cboettig:
Closes #420.

## What changed

| Tool | Before | After |
|---|---|---|
| `get_stac_details` | `dataset_id, catalog_url?, catalog_token?, collection?` | `dataset_id, collection?` |
| `get_collection` | `collection_id, catalog_url?, catalog_token?, collection?` | `collection_id, collection?` |
| `browse_stac_catalog` | `catalog_url?, catalog_token?, catalog?` | `catalog?` — and registered only when `STAC_DISCOVERY` allows it |

Two new deployment knobs replace what left the schema:

| Var | Default | Meaning |
|---|---|---|
| `STAC_CATALOG_TOKEN` | unset | Bearer token for `STAC_CATALOG_URL` when the deployment's catalog is private. Sent on catalog fetches (startup pre-warm, periodic refresh, cache-miss refetch); never in a JSON-RPC body. |
| `STAC_DISCOVERY` | `1` | Whether to register `browse_stac_catalog` + `catalog://list`. `0`/`false`/`off` leaves both out of `tools/list`. |

## Why this is safe

Inline `collection` already wins the resolution order, and geo-agent always sends it (`map-tools.js` `get_schema`, the `injectInlineStac` argsRewriter). So the parameters were only ever reachable by the model, for the case #420 exists to close.

**A client still sending them is not broken.** MCP argument validation drops fields absent from the schema, so the call succeeds — the value is simply ignored and the lookup resolves against the deployment's own catalog. That is the intended new behavior, which answers one of the open questions: no deprecation window is needed, the tolerance is structural. It is covered by a test rather than asserted.

The wyoming sidecar's `_inject()` is the one live injector of `catalog_url`, and it lands in that same ignore path.

## Open questions, as resolved here

- **Discovery default on or off** — on, per your comment. `utah-public-lands` is third-party, pinned at geo-agent `v3.24.0`, pointed at the shared public server whose configured catalog *is* the public catalog; defaulting off would break its prompt with no way for us to fix it. Private apps run their own replica and set `STAC_DISCOVERY=0`.
- **Deprecation window** — not needed, see above.
- **`catalog://` resources** — `catalog://list` carries the same whole-catalog reliance as `browse`, so it is gated with it. `catalog://{dataset_id}` is a per-dataset lookup scoped to the deployment's catalog and stays registered either way.
- **Fold the rule back into the architecture doc** — done: `docs/architecture/catalog-sourcing.md` gains a *"scope is ambient, never an argument"* section, so the next tool with a convenient `catalog_url` param has something to be pointed at.

## Guidance

`STAC_DISCOVERY=0` also drops the "call `browse_stac_catalog`" step from the CRITICAL block and from the `query` description — guidance naming a tool the deployment does not register just burns a turn on `Unknown tool`. **With discovery on, both strings are byte-identical to what shipped before**: the assembled `query` description measures 40,969 ch on this branch and on `main`. So the open-model matrix is not gated on this for dev/prod; it would only be needed before a deployment actually flips the flag.

One incidental fix: `get_dataset`'s not-found message said *"Use list_datasets"*, which is not a tool name. It now names `browse_stac_catalog`, and says nothing when discovery is off.

## Verification

The issue's five checks, as tests (`tests/test_server.py::TestCatalogScopeIsNotAToolArgument`, `::TestDiscoveryIsOptOutPerDeployment`, `tests/test_stac.py::TestDeploymentCatalogScope`, `::TestDiscoveryFlag`):

1. ✅ `tools/list` advertises neither parameter on any tool — asserted on both the signatures and the registered input schemas.
2. ✅ `get_stac_details` with inline `collection` returns what it did before.
3. ✅ With discovery disabled, `browse_stac_catalog` is absent from `tools/list` (and `catalog://list` from `resources/list`).
4. ✅ A call naming a foreign `catalog_url` does not fetch it — the test fails the call if `pystac.Catalog.from_file` or `fetch_stac_catalog` is reached.
5. Wyoming `gye-parcels` — needs the replica, see below.

Full suite green locally: 142 + 124 + 186 passed. Guidance gate passes at 40,969 / 42,500 ch.

## Rollout note

Not done here, and worth a separate issue on the app repo: **wyoming's replica must set `STAC_CATALOG_URL` to its own in-cluster catalog (plus `STAC_CATALOG_TOKEN`) and `STAC_DISCOVERY=0` before it stops relying on per-request `catalog_url`.** Today it points at the public NRP catalog and reaches the private one per call, so the app holding confidential parcel data has the widest discovery surface of any deployment. Its configured-data lookups keep working through inline `collection` regardless — but the fix that #420 promises it only lands when the replica's env moves. That also makes the sidecar's `catalog_url` rewrite dead code.

No change to `k8s/deployment.yaml` or `k8s/dev-deployment.yaml`: both want the default (discovery on, public catalog, no token).


cboettig:
Added in 5998b37, per review discussion.

The silent-ignore property is good for the known consumers and bad for an unknown one: geo-agent doesn't send `catalog_url` and the wyoming sidecar doesn't need its copy honoured, but a direct MCP consumer that genuinely relied on it is now answered from the deployment's catalog with no error — wrong data that reads exactly like right data. The log line #420 suggested isn't available to us: validation drops the argument before any of our code runs.

So the signal goes in the response that consumer will actually read. On a miss:

```
Dataset 'ca-parcels' not found in this server's catalog
(https://s3-west.nrp-nautilus.io/public-data/stac/catalog.json).
Use browse_stac_catalog to list available datasets.
```

`browse_stac_catalog` already named its catalog in the header; `get_stac_details` and `get_collection` did not. The URL quoted is the client-reachable one (#346) — on a mirror head the address this process reads may not resolve for the caller, and a miss that quotes an unfetchable URL is not diagnosable. Two tests cover that and the plain case.

Worth noting for anyone reading the verification list: this does not make a dropped `catalog_url` *detectable* on a hit — a lookup that succeeds against the deployment's catalog still looks ordinary. It makes the failure mode legible, which is where a confused consumer ends up.

No guidance movement: these are runtime tool responses, not injected description. Still 40,969 ch.

